Issue #92, opened 2023-07-22
npm run start:dev lists @types typescript issues


negue:
Not a bug of the app, I wanted to try to compile it but somehow the installed @types are invalid?

```
ERROR in /Users/negue/GitHub/anytype-ts/node_modules/@types/yargs/index.d.ts
[tsl] ERROR in /Users/negue/GitHub/anytype-ts/node_modules/@types/yargs/index.d.ts(855,29)
      TS1005: ';' expected.

ERROR in /Users/negue/GitHub/anytype-ts/node_modules/@types/yargs/index.d.ts
[tsl] ERROR in /Users/negue/GitHub/anytype-ts/node_modules/@types/yargs/index.d.ts(925,79)
      TS1005: ';' expected.

ERROR in /Users/negue/GitHub/anytype-ts/node_modules/@types/yargs/index.d.ts
[tsl] ERROR in /Users/negue/GitHub/anytype-ts/node_modules/@types/yargs/index.d.ts(945,1)
      TS1160: Unterminated template literal.

webpack 5.88.2 compiled with 1050 errors in 23526 ms
```

Looking at those files the normal typescript highlighting / parser doesn't mark anything bad, so I wonder what is causing these issues... am I missing something? 😬 

OS: Mac Ventura 13.4 M1
node -v: v16.18.1  (too old version?)
npm -v: 9.1.2

ra3orblade:
Seems that your typescript version is incorrect.

negue:
I always thought during these compile steps it would use the locally typescript version that is set in the package.json, which is `^3.9.7` I checked in the node_modules/typescript Folder and the installed one is ` "version": "3.9.10",` so still in the same range.

Shouldn't be this installed local version be used inside the webpack build step?

Just checked my globally installed tsc - it was 4.1.3 updated that one to the newest - but its the same issue. 
Or do I need to downgrade my global tsc to let it compile ?

negue:
Installing typescript `3.9.7` globally also results in the same error - so neither an upgrade/downgrade globally helped

negue:
After countless reinstalls of everything, I tried the crazy idea to just install the packages using `npm ci` install `npm install` and those TypeScript "issues" (since the d.ts files looked good) are GONE! 🎉 

I assume that all packages are not pinned (to the exact version) the `npm install` wanted to be smart and install the updated version of the rules `^` and I guess some sub module of a sub module broke it completely.  

If you want I could create a PR to pin those.

---

[...]

PR #94, opened 2023-07-22
chore: simplify update script

cmars:
anyproto-heart artifacts do not need github authentication to download. This might be left over from before this project was public. Anonymous downloading is safer for developers and preserves a bit more privacy.

Make "ubuntu-latest" the default platform. A bit opinionated, but linux is generally a safe default for CI and Docker, so might be reasonable. I can yoink this particular if it's too controversial.

Thanks!

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -4,10 +4,8 @@ REPO="anyproto/anytype-heart"
 FILE="addon.tar.gz"
 GITHUB="api.github.com"
 
-user=$1
-token=$2;
-platform=$3;
-arch=$4;
+platform=${1:-ubuntu-latest};
+arch=$2;
 folder="build";
 
 if [ "$platform" = "ubuntu-latest" ]; then
@@ -26,11 +24,6 @@ echo "Arch: $arch"
 echo "Folder: $folder"
 echo ""
 
-if [ "$token" = "" ]; then
-  echo "ERROR: token is empty"
-  exit 1
-fi;
-
 if [ "$arch" = "" ]; then
   echo "ERROR: arch not found"
   exit 1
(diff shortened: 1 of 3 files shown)


ra3orblade:
You have changed update.sh but it's part of CI process and ubuntu-latest is a platform used by Github actions, so I can't accept this change unfortunately.

cmars:
> You have changed update.sh but it's part of CI process and ubuntu-latest is a platform used by Github actions, so I can't accept this change unfortunately.

Ah, makes sense! I've reverted that bit. Thanks for reviewing!

cmars:
Updated github actions with the update.sh argument changes.

negue:
Oh nice - I also wanted to ask about the Auth not needed anymore in the update.sh 🎉 

@cmars did you had no issues while trying to `npm run start:dev` ?  I had to push the `anytypeHelper` into the dist folder first  ( #92 ) - so I wonder if thats only an error on my side   or if the update.sh needs another step to copy the binary into dist

ra3orblade:
> Oh nice - I also wanted to ask about the Auth not needed anymore in the update.sh 🎉
> 
> @cmars did you had no issues while trying to `npm run start:dev` ? I had to push the `anytypeHelper` into the dist folder first ( #92 ) - so I wonder if thats only an error on my side or if the update.sh needs another step to copy the binary into dist

As it runs two times for macos and linux it basically creates two folders darwin-amd and darwin-arm which are then copied to dist by electronn hook. So no, update.sh doesnt need this step, if you want update.sh to work correctly you can just fix README.md, imo, which adds `mv anytypeHelper` step, do not forget that windows has exe file.

ra3orblade:
> Updated github actions with the update.sh argument changes.

looks better, I will check once again and merge when I will be able to merge, right now my colleague is doing some licensing magic =)

update.sh
```diff
@@ -4,10 +4,8 @@ REPO="anyproto/anytype-heart"
 FILE="addon.tar.gz"
 GITHUB="api.github.com"
 
-user=$1
-token=$2;
-platform=$3;
-arch=$4;
+platform=${1:-ubuntu-latest};
```
ra3orblade, line 7: is it a default value? should not be ${1:=ubuntu-latest} ?

cmars: Well.. we could use that form, but it'd then be more like:

```
platform=$1
: ${platform:=ubuntu-latest}
```

Ref: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

ra3orblade: oh, ok sorry, ima noob in bash

cmars: No worries! I got curious and had to look up `:=` myself :)

fuksman:
recheck

cmars:

PR #99, opened 2023-07-23
Extend Readme to copy anytypeHelper into dist

negue:
As mentioned in #94 and #92 this adds the information that the binary needs to be moved when running it locally.


negue:

ra3orblade:
Add commands for all platforms pls

negue:
Added a table overview of the commands and updated the gitignore because the `update.sh` also accepts `arm`/ `amd` without `64`

I would've added the commands to move them from the `anytype-heart` too, but I couldn't get the `make install-dev-js` command to work :/

ra3orblade (approved):

ra3orblade:
> Added a table overview of the commands and updated the gitignore because the `update.sh` also accepts `arm`/ `amd` without `64`
> 
> I would've added the commands to move them from the `anytype-heart` too, but I couldn't get the `make install-dev-js` command to work :/

thanks, merged

PR #106, opened 2023-07-25
Improve linting

### LINKED ISSUE #78: [RFC] Adopt code formatters?
Even though prettier formatter seems to be used by some developers[^1], usage of it does not appear to be enforced (i.e. not part of the linting process). Heck, even the prettier config file itself isn't formatted!

I would like to open some PRs listed below, but I'm not sure if that's something anytype developers are happy with. So I'm asking for confirmation here. Also, this RFC isn't concerned with non-prettifyable files such as shell scripts.

Judging by [`js-anytype.code-workspace`](https://github.com/anyproto/anytype-ts/blob/ee5f69e7d062b0dac1c0d9807183171bc08e838b/js-anytype.code-workspace) and [`.gitignore`](https://github.com/anyproto/anytype-ts/blob/ee5f69e7d062b0dac1c0d9807183171bc08e838b/.gitignore) file, IDEs used by anytype developers are vscode, eclipse, webstorm/intelliJ so I will open PRs based on that assumption (if confirmed).

## PR 1 : Add prettier formatter

- update documentations for setting up formatters for different IDEs
- add eslint husky hook
- remove following lines
  https://github.com/anyproto/anytype-ts/blob/ee5f69e7d062b0dac1c0d9807183171bc08e838b/js-anytype.code-workspace#L18-L24

## PR 2 : Format files affected by prettier

- Simply apply prettier formatting to affected files
- Is done to make diffs more managable and no malicious code slips through the review process
- Anytype developers can verify that no other alterations has been made

[^1]: https://github.com/anyproto/anytype-ts/blob/ee5f69e7d062b0dac1c0d9807183171bc08e838b/.prettierrc.json

developomp:
- removes prettier formatter
- removes deprecated tslint
- fixes eslint errors

## Note

- eslint still has bunch of warnings and suppressed errors
  - check `.eslintrc.js` to see suppressed errors
  - run `npm run lint` to see all warnings
  - the goal is to fix them one by one at a later point in time
    - this is done to keep the PR as simple as possible


ra3orblade:
I'm sorry but I can't accept PR that changes whole codebase. Another thing is that it still has problems (these are the things i've found at first glance):
- Tabs are changed to spaces
- Quotes are changed to double quotes

developomp:
> can't accept PR that changes whole codebase

99% of the changes come from the very last commit which can be easily verified. Also it's kinda inevitable when adopting code formatter.

> Tabs are changed to spaces

I thought you were [fine with either](https://github.com/anyproto/anytype-ts/issues/78#issuecomment-1651216146) but I'll change that.

> Quotes are changed to double quotes (i.e. in imports), double quotes are only used in JSON and JSX/HTML attributes

This I can also change

> Removed spaces in braces

Is this a deal-breaker? If not, I think the consistency and developer experience improvement for the contributors is very worth the sacrifice.

ra3orblade:
I think that code formatting first of all should be applied to new code and not change every line in every file, this is not acceptable, and yes it should follow guidelines of the current code.

developomp:
Should be fine now.

developomp:
Should we just get rid of code formatters entirely and just use eslint? I think that should be good enough for most people anyway.

developomp:
> it should follow guidelines of the current code.

Maybe we can work on a proper code guideline sometime in the future.

developomp:
@ra3orblade Sorry for causing all the headache. I have removed everything related to prettier formatting and reduced the scope of this PR to affect eslint only.

ra3orblade (approved):

PR #107, opened 2023-07-25
Fix lint-staged not applying to any file

developomp:
Running `npm run precommit` doesn't show any error, but that was because it wasn't being run on any files in the first place. This PR fixes that. The `tsc` stuff was removed because of this (https://github.com/okonet/lint-staged/issues/468) and because eslint handles typescript errors for us anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -550,8 +550,7 @@
 		}
 	},
 	"lint-staged": {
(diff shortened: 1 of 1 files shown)


developomp:
@ra3orblade review please! :D

ra3orblade (approved):

PR #118, opened 2023-07-27
Automatically copy files after running update.sh

developomp:


ra3orblade:
This is incorrect, Linux and MacOS builds run twice for AMD and ARM architecture and update.sh downloaded folders are then moved by after build hook in electron-builder.

developomp:
That makes sense. Thank you for explaining!

PR #119, opened 2023-07-27
Tidy up dist/lib gitignore

developomp:
- cleaner gitignore structure
- prvent files from showing up on git status after running `update.sh`


ra3orblade:
This is incorrect, directory structure should stay intact, your PR removes lib entirely.

developomp:
I think it's pretty safe to delete them. They get repopulated after running the `update.sh` script anyway. I don't see the point in keeping them.

ra3orblade:
They are now used in localisation update mechanism.

developomp:
My main concern is that every time someone runs `update.sh` or `npm run start:dev`, it makes `git status` messy.

ra3orblade:
> My main concern is that every time someone runs `update.sh` or `npm run start:dev`, it makes `git status` messy.

could you show example?

developomp:
image

ra3orblade:
`https://github.com/anyproto/anytype-ts/commit/143e751ab07cbe98028a8e759620db1b445dddca` this fixes the above screen shot, but actually if you want to make command for updating without changing the update.sh is just `update.sh && git checkout . && git clean -df`

developomp:
Verified that 143e751ab07cbe98028a8e759620db1b445dddca fixed the issue. Thanks!

PR #238, opened 2023-08-30
Filter by type on collection doesn't work properly

Shampra:
---

- [X] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

### Description
So, when user filter by object type, he can choose in a list.
But if this type is not already present as linked object to the collection :
- filter seems to be applied with the type added (type doesn't appear in the list anymore, confirmed by test in dev)
- type doesn't appears in the "Object type" in list of active filter 
- type doesn't appears in selected type list when we edit this filter  (so, it cannot be removed)

Object was return with  detail.get(subId, id), and for collection type returned are empty.

Pehaps, there is a reason of this (like not allowing a collection to be filtered if the type is not in the collection)? If yes, the list of selection in search must be filtered too.
I suppose no (I can filter by type=audio, if I don't have one I won't get any results and that's it).
It's applied via this PL, I hope it's good for you :)

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://community.anytype.io/t/object-type-filter-is-not-working-properly-in-collections/10768

_Collection (with only page) filtered by type = set or Image or page, etc_
image
image

### Added tests?

- [ ] 👍 yes
- [X] 🙋 no, because I need help

### Added to documentation?

- [ ] 📜 README.md
- [X] 🙅 no documentation needed

### DIFF
--- a/src/ts/component/menu/dataview/filter/values.tsx
+++ b/src/ts/component/menu/dataview/filter/values.tsx
@@ -138,7 +138,7 @@ const MenuDataviewFilterValues = observer(class MenuDataviewFilterValues extends
 					);
 				};
 
-				list = Relation.getArrayValue(item.value).map(it => detailStore.get(subId, it, []));
+				list = Relation.getArrayValue(item.value).map(it => dbStore.getType(it));
 				list = list.filter(it => !it._empty_);
 
 				value = (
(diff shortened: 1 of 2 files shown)


ra3orblade:
This is not correct, filter can contain other objects, not only types. This is bug on middleware and should be fixed there.

Issue #395, opened 2023-12-04
Auto indenting/closing braces while using a code block


NextPr0gram:
### Have you read a contributing guide?

- [X] I have read CONTRIBUTING.md
- [X] I have searched the existing requests and didn't find any that were similar
- [X] I have considered creating a pull request instead and want to proceed

### Clear and concise description of the problem

When entering the following bracket types: "(", "[", "{", they should automatically close and the cursor should be placed in between the brackets. 
if the "Enter" key is pressed when the cursor is in between brackets then the cursore will be positioned on the next line and will be indented, and the closing bracket will be positioned on the next line after that.
EXAMPLE:

function = {|}

press enter key

function = {
    |
}

LavaCxx:
I submitted a PR for the first part. 
However, there are quite a lot of code logic that needs to be considered when implementing the second part, and the complexity would be a bit high for a code block that only serves demonstration purposes. Therefore, similar software rarely achieves the same feature.

PR #476, opened 2024-01-10
Apply auto close braces in code block

LavaCxx:
---

---

### Description
To provide some convenience for directly writing code in code blocks.
Also provided some additional support for full-width symbols, such as `【「(“‘` etc.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

#395 

demo

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/block/text.tsx
+++ b/src/ts/component/block/text.tsx
@@ -703,7 +703,7 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 			};
 		};
 		[];
-		const twineOpen = [ '[', '{', '\'', '\"', '(', '【','「','（','“','‘' ];
+		const twineOpen = [ '[', '{', '\'', '\"', '`', '(', '【', '「', '（', '“', '‘' ];
 		const twineClose = {
 			'[': ']',
 			'{': '}',
--- a/src/ts/component/block/text.tsx
+++ b/src/ts/component/block/text.tsx
@@ -643,7 +643,7 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 		e.persist();
 
 		const { onKeyDown, rootId, block, isInsideTable } = this.props;
-		const { id } = block;
+		const { id,content:{text} } = block;
 
 		if (menuStore.isOpenList([ 'blockStyle', 'blockColor', 'blockBackground', 'blockMore' ])) {
 			e.preventDefault();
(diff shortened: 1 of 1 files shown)


src/ts/component/block/text.tsx
```diff
@@ -643,7 +643,7 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 		e.persist();
 
 		const { onKeyDown, rootId, block, isInsideTable } = this.props;
-		const { id } = block;
+		const { id,content:{text} } = block;
```
ra3orblade, line 646: I do not see where it's used and we do not write this way, `const { text } = block.content;` for example.

src/ts/component/block/text.tsx
```diff
@@ -702,12 +702,17 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 				saveKeys.push({ key: `arrowright, arrowdown` });
 			};
 		};
-
-		const twineOpen = [ '[', '{', '\'', '\"', '(' ];
+		[];
```
ra3orblade, line 705: Useless braces?

src/ts/component/block/text.tsx
```diff
@@ -702,12 +702,17 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
...
+		const twineOpen = [ '[', '{', '\'', '\"', '`', '(', '【', '「', '（', '“', '‘' ];
```
ra3orblade, line 706: This needs refactoring, we can leave only one map and use it's keys as opening symbols.

src/ts/component/block/text.tsx
```diff
@@ -859,7 +863,7 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 				focus.set(block.id, { from: range.from + l, to: range.to + l });
 				focus.apply();
 			});
-
+			{}
```
ra3orblade, line 866: Unused braces?

ra3orblade (changes requested):

src/ts/component/block/text.tsx
LavaCxx: Sorry, I forgot to delete some test codes. The original intention was to try automatic indentation.

src/ts/component/block/text.tsx
LavaCxx: I tested how vscode handles indentation on the same page, but I forgot to delete it, which was a mistake that should not have happened.

src/ts/component/block/text.tsx
```diff
@@ -843,13 +850,12 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 			e.preventDefault();
 			this.onSmile();
 		});
-
-		if (range && (range.from != range.to) && twineOpen.includes(key)) {
+		if (range && ((range.from != range.to) || block.isTextCode()) && twinePairs[key]) {
```
ra3orblade, line 853: This will work with a closing symbol as well, `Object.keys(twinePairs).includes(key)` I meant

ra3orblade (changes requested):

src/ts/component/block/text.tsx
LavaCxx: How about `Reflect.has(twinePairs,key)`

ra3orblade: Ok, never used it though)

ra3orblade (approved):

PR #515, opened 2024-01-24
Prohibit clicking already activated tab

LavaCxx:
---

---

### Description
CLS are very annoying things, and would be even more annoying if they could frequently trigger.
It is better to start from prohibiting activated tab from being repeatedly clicked.
before

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [x] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/page/main/store.tsx
+++ b/src/ts/component/page/main/store.tsx
@@ -322,8 +322,12 @@ const PageMainStore = observer(class PageMainStore extends React.Component<I.Pag
 	onTab (id: any, isInner: boolean) {
 		const { isPopup } = this.props;
 
+		if (this.tab == id) {
+			return;
+		};
+
 		this.tab = id;
-		this.onView(Storage.get('viewStore') || View.Library, isInner);
+		this.onView(Storage.get('viewStore') || View.Library, isInner, true);
 
 		Storage.set('tabStore', id);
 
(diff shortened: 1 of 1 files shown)


ra3orblade:
I think it's better to rework these flags as force flag or initially set tab as null to avoid this onMount logic.

LavaCxx:
> I think it's better to rework these flags as force flag or initially set tab as null to avoid this onMount logic.

Modified.

src/ts/component/page/main/store.tsx
```diff
@@ -340,7 +344,11 @@ const PageMainStore = observer(class PageMainStore extends React.Component<I.Pag
 		};
 	};
 
-	onView (id: View, isInner: boolean) {
+	onView (id: View, isInner: boolean, isChangeTab: boolean = false) {
+		if (!isChangeTab && this.view == id) {
```
ra3orblade, line 348: add parenthesis please `&& (this.view == id)`

ra3orblade (changes requested):

ra3orblade (approved):

fuksman:
@any contributor @LavaCxx code

PR #520, opened 2024-01-25
Bilibili embed set default autoplay

LavaCxx:
---

---

### Description

Unlike YouTube, Bilibili videos automatically play by default when `autoplay` value is unset.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/block/embed.tsx
+++ b/src/ts/component/block/embed.tsx
@@ -645,9 +645,10 @@ const BlockEmbed = observer(class BlockEmbed extends React.Component<I.BlockComp
 					if (block.isEmbedBilibili() && text.match(/src="\/\/player[^"]+"/)) {
 						text = text.replace(/src="(\/\/player[^"]+)"/, 'src="https:$1"');
 
-						if (!/autoplay=/.test(text)) {
-							text = text.replace(/(src="[^"]+)"/, `$1&autoplay=0"`);
-						};
+					};
+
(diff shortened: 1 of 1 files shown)


src/ts/component/block/embed.tsx
```diff
@@ -644,6 +644,10 @@ const BlockEmbed = observer(class BlockEmbed extends React.Component<I.BlockComp
 					// Fix Bilibili schemeless urls
 					if (block.isEmbedBilibili() && text.match(/src="\/\/player[^"]+"/)) {
```
ra3orblade, line 645: but it will change autoplay only if the link is incorrect this way

ra3orblade (changes requested):

src/ts/component/block/embed.tsx
LavaCxx: The URL copied from the website simply lacks https:, which generally is not a problem. But it's still worth considering this situation.

ra3orblade (approved):

PR #522, opened 2024-01-28
Fix Form input zero values:

Soleone:
### Description

Without this fix when entering 0 as a number to save in an input field it will instead not save anything and treat it as null

- [x] 🐛 Bug Fix

AnytypeInputZeroBug

### Added tests?

- [ ] 👍 yes
- [x] 👎  no, because I couldn't find any existing tests

### Added to documentation?

### DIFF
--- a/src/ts/component/form/input.tsx
+++ b/src/ts/component/form/input.tsx
@@ -226,7 +226,7 @@ class Input extends React.Component<Props, State> {
 			return;
 		};
 
-		this.state.value = String(v || '');
+		this.state.value = String(v ?? '');
 		this.setState({ value: this.state.value });
 	};
 	


Soleone:
recheck

src/ts/component/form/input.tsx
```diff
@@ -226,7 +226,7 @@ class Input extends React.Component<Props, State> {
 			return;
 		};
 
-		this.state.value = String(v || '');
```
Soleone, line 229: `||` treats `0` as falsy, similar to `null` which leads to just `''` being stored here.

Soleone:

Soleone:
recheck

ra3orblade (approved):

fuksman:
@any contributor @Soleone code

PR #594, opened 2024-03-03
fix: scrollbar cls

LavaCxx:
---

---

### Description
Fix the CLS issue in the editor caused by the scroll bar.

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [x] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

[Unexpected sidebar popup](https://community.anytype.io/t/unexpected-sidebar-popup/16916)

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/scss/page/main/edit.scss
+++ b/src/scss/page/main/edit.scss
@@ -1,6 +1,6 @@
 @import "~scss/_vars";
 
 .pageMainEdit {
-	#bodyWrapper { padding: 52px 0px 0px 0px; }
+	#bodyWrapper { padding: 52px 0px 0px 0px; overflow: overlay; scrollbar-gutter: stable both-edges; }
 	#bodyWrapper > #loader { position: fixed; top: 0px; width: 100%; height: 100%; }
 }


ra3orblade:
I didn't understand what you are trying to do here - the editor scrollbar is not overflown, it's just window scrollbar. If you were trying to fix page moving when opening a popup it should be added to body.over as this class is added when popup is opened.

LavaCxx:
> I didn't understand what you are trying to do here - the editor scrollbar is not overflown, it's just window scrollbar. If you were trying to fix page moving when opening a popup it should be added to body.over as this class is added when popup is opened.
> 
> UPDATE: I've read the forum post but still do not understand why it should fix it as the scrollbar is not from #bodyWrapper.

`scrollbar-gutter` should be retained at least, as it can help avoid the jitter caused by the scroll bar.

ra3orblade:
> > I didn't understand what you are trying to do here - the editor scrollbar is not overflown, it's just window scrollbar. If you were trying to fix page moving when opening a popup it should be added to body.over as this class is added when popup is opened.
> > UPDATE: I've read the forum post but still do not understand why it should fix it as the scrollbar is not from #bodyWrapper.
> 
> `scrollbar-gutter` should be retained at least, as it can help avoid the jitter caused by the scroll bar.

Maybe, but I've tried to add it to the class .overPopup and it does not work when I turn on "Always show scrollbars" in MacOS. I can merge it but just wondering.

ra3orblade (approved):

PR #695, opened 2024-04-27
Last opened object homepage setting with window context

mikailcf:

---

### What

This PR makes it so that the _Last opened object_ homepage setting now considers the context of different windows (i.e. different windows have different last opened objects and each is loaded accordingly when the _Last opened object_ home page setting is selected).

Also when closing a window its object is removed from the local storage _Last opened_ context and when closing the app all objects, besides the last opened for by the first window, are removed from the local storage.

### Why

*The reasoning here is based on my usage and is open for discussion*

#### Practical reason

Currently when users have the _Last opened object_ homepage selected and have multiple windows opened, any time any of those windows are refreshed it then loads the last opened object period (which in many cases was opened in another window).

One example in my workflow is: I have a window with a set of active notes on different topics and frequently either create a new note in the set and open it in a new window or open an active note in a new window. The window with the set of notes should always only display that object, but opening new notes on new pages overrides the _Lsat opened object_ and after having the computer go to sleep and wake it up again, now all windows show the same object (which was the created/opened note).

#### Reasonable... reason

It seems reasonable to me that users with the _Last opened object_ homepage selected wouldn't want all windows to load the same object after a mass refresh (such as the computer waking up from sleep) 😜 .

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/electron/js/window.js
+++ b/electron/js/window.js
@@ -226,11 +226,6 @@ class WindowManager {
 		return { x, y };
 	};
 
-	getWindowId () {
-		const currentWindow = BrowserWindow.getFocusedWindow();
-		return currentWindow.id;
-	}
-
 	sendToAll () {
 		const args = [ ...arguments ];
 		this.list.forEach(it => Util.send.apply(this, [ it ].concat(args)));
--- a/electron/js/window.js
+++ b/electron/js/window.js
@@ -53,6 +53,10 @@ class WindowManager {
 
 		this.list.add(win);
 
+		win.on('close', () => {
+			Util.send(win, 'will-close-window', win.id);
+		});
+
 		win.on('closed', () => {
 			this.list.delete(win);
 			win = null;
@@ -60,7 +64,7 @@ class WindowManager {
 
 		win.on('focus', () => { 
 			UpdateManager.setWindow(win);
-			MenuManager.setWindow(win); 
+			MenuManager.setWindow(win);
 		});
 		win.on('enter-full-screen', () => Util.send(win, 'enter-full-screen'));
 		win.on('leave-full-screen', () => Util.send(win, 'leave-full-screen'));
@@ -234,7 +238,7 @@ class WindowManager {
 	reloadAll () {
 		this.sendToAll('reload');
 	};
-
+	
 };
 
 module.exports = new WindowManager();
--- a/electron/js/window.js
+++ b/electron/js/window.js
@@ -226,6 +226,11 @@ class WindowManager {
 		return { x, y };
 	};
 
+	getWindowId () {
+		const currentWindow = BrowserWindow.getFocusedWindow();
+		return currentWindow.id;
+	}
+
 	sendToAll () {
 		const args = [ ...arguments ];
 		this.list.forEach(it => Util.send.apply(this, [ it ].concat(args)));
--- a/src/ts/app.tsx
+++ b/src/ts/app.tsx
@@ -279,10 +279,12 @@ class App extends React.Component<object, State> {
 		Renderer.on('leave-full-screen', () => commonStore.fullscreenSet(false));
 		Renderer.on('logout', () => authStore.logout(false, false));
 		Renderer.on('data-path', (e: any, p: string) => commonStore.dataPathSet(p));
+		Renderer.on('will-close-window', this.onWillCloseWindow);
 
 		Renderer.on('shutdownStart', () => {
 			this.setState({ loading: true });
 			Storage.delete('menuSearchText');
+			Storage.deleteAllLastOpened();
 		});
 
 		Renderer.on('zoom', () => {
@@ -392,6 +394,10 @@ class App extends React.Component<object, State> {
 		} else {
 			cb();
 		};
+	};
+
+	onWillCloseWindow (e: any, windowId: string) {
+		Storage.deleteLastOpenedByWindowId([windowId]);
 	};
 
 	onPopup (e: any, id: string, param: any, close?: boolean) {
--- a/src/ts/lib/api/command.ts
+++ b/src/ts/lib/api/command.ts
@@ -1309,9 +1309,8 @@ export const ObjectOpen = (objectId: string, traceId: string, spaceId: string, c
 		// Save last opened object
 		const object = detailStore.get(objectId, objectId, []);
 		const windowId = UtilCommon.getCurrentElectronWindowId();
-
 		if (!object._empty_ && ![ I.ObjectLayout.Dashboard ].includes(object.layout)) {
-			Storage.set('lastOpened', { [windowId]: { id: object.id, layout: object.layout, spaceId: object.spaceId }});
+			Storage.setLastOpened(windowId, { id: object.id, layout: object.layout, spaceId: object.spaceId });
 		};
 
 		if (callBack) {
--- a/src/ts/lib/api/command.ts
+++ b/src/ts/lib/api/command.ts
@@ -1308,8 +1308,10 @@ export const ObjectOpen = (objectId: string, traceId: string, spaceId: string, c
 
 		// Save last opened object
 		const object = detailStore.get(objectId, objectId, []);
+		const windowId = UtilCommon.getCurrentElectronWindowId();
+
 		if (!object._empty_ && ![ I.ObjectLayout.Dashboard ].includes(object.layout)) {
-			Storage.set('lastOpened', { id: object.id, layout: object.layout, spaceId: object.spaceId });
+			Storage.set('lastOpened', { [windowId]: { id: object.id, layout: object.layout, spaceId: object.spaceId }});
 		};
 
 		if (callBack) {
--- a/src/ts/lib/storage.ts
+++ b/src/ts/lib/storage.ts
@@ -58,35 +58,6 @@ class Storage {
 			this.storage[key] = JSON.stringify(o);
 		};
 	};
-
-	unsetPath (key: string, path: string[]): void {
-		if (!key || !path) {
-			console.log('[Storage].unset: key or path not specified');
-			return;
-		};
-		
-		const o = this.get(key);
-
-		if ((typeof o !== 'object') || (o !== null) || path.length == 0) {
-			console.log('[Storage].unset: unset should be used on a stored object with a valid path');
-			return;
-		}
-
-		const lastKey = path.pop();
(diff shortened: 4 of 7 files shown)


electron/js/window.js
```diff
@@ -226,6 +226,11 @@ class WindowManager {
 		return { x, y };
 	};
 
+	getWindowId () {
```
mikailcf, line 229: This isn't currently used by I assume it will be when implementing the _remove from storage when closing window_ part. If it isn't I'll remove it.

src/ts/lib/api/command.ts
```diff
@@ -1308,8 +1308,10 @@ export const ObjectOpen = (objectId: string, traceId: string, spaceId: string, c
 
 		// Save last opened object
 		const object = detailStore.get(objectId, objectId, []);
+		const windowId = UtilCommon.getCurrentElectronWindowId();
+
 		if (!object._empty_ && ![ I.ObjectLayout.Dashboard ].includes(object.layout)) {
-			Storage.set('lastOpened', { id: object.id, layout: object.layout, spaceId: object.spaceId });
+			Storage.set('lastOpened', { [windowId]: { id: object.id, layout: object.layout, spaceId: object.spaceId }});
```
mikailcf, line 1314: I leverage the fact that the current implementation of `set` merges objects to have a minimally intrusive change.

src/ts/lib/storage.ts
```diff
@@ -58,6 +58,35 @@ class Storage {
 			this.storage[key] = JSON.stringify(o);
 		};
 	};
+
+	unsetPath (key: string, path: string[]): void {
```
mikailcf, line 62: This is used to remove a key at a path in a stored object (in the case of this PR remove the last opened object indexed by the window id).

mikailcf:
I hereby grant Any Association and its successors or assignees permission to use, copy, modify, distribute, and sub-license my contributions on any terms they like, under all my rights in my contributions. A contribution includes anything I submit to the project. I am giving them this license in order to make it possible for them to accept my contributions into their project.

As far as the law allows, my contributions come as is, without any warranty or condition, and I will not be liable to anyone for any damages related to this software or this license, under any kind of legal claim.

mikailcf:
recheck

mikailcf:

mikailcf:
recheck

ra3orblade:
I do not like that you have changed Storage.set, it should not contain this logic. Storage.set should not know any context about the key that is being set and you have incorporated this logic inside Storage.set.

All helping method that store objects should be separate, like Storage.setToggle, Storage.setScroll and etc.

Second thing - check formatting and linting please.

mikailcf:
Do you mean the change I made when calling `Storage.set`?

The responsibility of knowing the context is in the caller (`ObjectOpen` function) in this case. The implementation of `Storage.set` is unchanged.

---
I fixed the linting. If you have any other style requests I'd ask you to comment in the specific lines, please.

ra3orblade:
Sorry, I've mixed the method when I was writing and was looking at unsetPath.

I mean the whole logic can work the same way it works in other helpers.

```
setLastOpened (windowId: string, param: any) {
	const obj = this.get('lastOpened') || {};
	obj[windowId] = Object.assign(obj[windowId] || {}, param);
	this.set('lastOpened', obj, true);
};

getLastOpened (windowId: string) {
	const obj = this.get('lastOpened') || {};
	return obj[windowId];
};
```

mikailcf:
Oh I see now, good idea. I've committed a new implementation following that.

src/ts/lib/storage.ts
```diff
@@ -108,6 +108,32 @@ class Storage {
...
+	unsetLastOpened (objectIds: string[]) {
+		const obj = this.get('lastOpened') || {};
+
+		let windowIdsToDelete = Object.keys(obj).reduce(
+			(windowIds, windowId) => objectIds.includes(obj[windowId].id) ? windowIds.concat(windowId) : windowIds, []
+		);
+
+		if (windowIdsToDelete.length == 0) {
+			return;
+		}
```
ra3orblade, line 126: Add semicolons

src/ts/lib/storage.ts
```diff
@@ -108,6 +108,32 @@ class Storage {
...
+		const obj = this.get('lastOpened') || {};
+		obj[windowId] = Object.assign(obj[windowId] || {}, param);
+		this.set('lastOpened', obj, true);
+	};
+
+	unsetLastOpened (objectIds: string[]) {
+		const obj = this.get('lastOpened') || {};
+
+		let windowIdsToDelete = Object.keys(obj).reduce(
+			(windowIds, windowId) => objectIds.includes(obj[windowId].id) ? windowIds.concat(windowId) : windowIds, []
```
ra3orblade, line 121: obj[windowId] can be null

src/ts/lib/util/common.ts
```diff
@@ -14,6 +14,16 @@ class UtilCommon {
 		return window.Electron || {};
 	};
 
+	getCurrentElectronWindowId (): string {
+		const electron = this.getElectron();
+
+		if (!electron) {
+			return '0';
+		}
```
ra3orblade, line 22: Fix semicolons

ra3orblade (changes requested):

arieldossantos (approved):
lgtm

alexjjassuncao (approved):

mikailcf:
@ra3orblade sorry about those approvals. I've mentioned that I had an open PR in an open source project to some coworkers and they got ahead of themselves and thought they could approve the changes. I've already dismissed their reviews.

---
About your requests: I'll push the necessary changes in a few hours.

src/ts/lib/storage.ts
```diff
@@ -108,6 +108,48 @@ class Storage {
...
+		};
+
+		windowIdsToDelete.forEach ((windowId) => delete(obj[windowId]));
+		this.set('lastOpened', obj, true);
+	};
+
+	deleteAllLastOpened () {
+		const obj = this.get('lastOpened') || {};
+		let windowIdsToDelete = Object.keys(obj);
+		this.deleteLastOpenedByWindowId(windowIdsToDelete);
```
ra3orblade, line 145: This is ambiguous,
```
this.set('lastOpened', {}, true);
```

src/ts/lib/storage.ts
```diff
@@ -108,6 +108,48 @@ class Storage {
...
+			return;
+		};
+
+		const obj = this.get('lastOpened') || {};
+
+		if (!homeIncluded) {
+			windowIdsToDelete = windowIdsToDelete.filter((id) => id !== '1');
+		};
+
+		windowIdsToDelete.forEach ((windowId) => delete(obj[windowId]));
```
ra3orblade, line 138: Wrong formatting, this logic is a bit unreadable
```
windowIdsToDelete.forEach(windowId => delete(obj[windowId]));
```

src/ts/app.tsx
```diff
@@ -394,6 +396,10 @@ class App extends React.Component<object, State> {
 		};
 	};
 
+	onWillCloseWindow (e: any, windowId: string) {
+		Storage.deleteLastOpenedByWindowId([windowId]);
```
ra3orblade, line 400: Wrong formatting:
```
[ windowId ]
```

src/ts/lib/util/common.ts
```diff
@@ -14,6 +14,16 @@ class UtilCommon {
...
 	};
 
+	getCurrentElectronWindowId (): string {
+		const electron = this.getElectron();
+
+		if (!electron) {
+			return '0';
+		};
+
+		return electron.currentWindow().windowId.toString();
```
ra3orblade, line 24: Why is it a string?

src/ts/lib/storage.ts
```diff
@@ -108,6 +108,48 @@ class Storage {
...
+
+	deleteLastOpenedByWindowId (windowIdsToDelete: string[], homeIncluded?: boolean) {
+		if (windowIdsToDelete.length == 0) {
+			return;
+		};
+
+		const obj = this.get('lastOpened') || {};
+
+		if (!homeIncluded) {
+			windowIdsToDelete = windowIdsToDelete.filter((id) => id !== '1');
```
ra3orblade, line 135: wrong formatting 
```
id => id != 1
```

src/ts/lib/storage.ts
```diff
@@ -108,6 +108,48 @@ class Storage {
...
+
+		let windowIdsToDelete = Object.keys(obj).reduce(
+			(windowIds, windowId) => !obj[windowId] || objectIds.includes(obj[windowId].id) ? windowIds.concat(windowId) : windowIds, []
+		);
+
+		this.deleteLastOpenedByWindowId(windowIdsToDelete, true);
+	};
+
+	deleteLastOpenedByWindowId (windowIdsToDelete: string[], homeIncluded?: boolean) {
+		if (windowIdsToDelete.length == 0) {
```
ra3orblade, line 128: No need to check length against zero
```
!windowIdsToDelete.length
```

ra3orblade (changes requested):

src/ts/lib/util/common.ts
mikailcf: I thought that making it a string might act a bit as a safeguard against `windowId` maybe changing types in the future and causing problems when using it as a key in the local storage.

src/ts/lib/storage.ts
mikailcf: `this.set('lastOpened', {}, true);` doesn't have the same effect. It removes all keys. The current call `deleteLastOpenedByWindowId(windowIdsToDelete)` keeps the `'1'` key intact.

src/ts/lib/storage.ts
mikailcf: Are you suggesting to change the stored type to number instead of string?

ra3orblade (approved):

fuksman:
@any contributor @mikailcf code

PR #709, opened 2024-05-04
Checkbox toggle shortcut

mikailcf:

---

### What
This PR adds a new shortcut for checking/unchecking checkboxes and suggests a change to the shortcut for expanding/collapsing toggles for both to be the same.

### Why
It's very practical to be able to mark checkbox items in a list when you can do that directly from the keyboard in a text heavy application.

The suggestion for the toggle shortcut is to keep both equal, since they both have similar effects, and with easier access. Since `cmd + enter` wasn't being currently used in Anytype and other similar apps in the market use that combination, I thought it would make sense.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -1129,7 +1129,7 @@
 	"popupShortcutNavigationMenu4": "Go to the right side of navigation. Link to page",
 	"popupShortcutNavigationMenu5": "Select option",
 	"popupShortcutNavigationPage": "Editor navigation",
-	"popupShortcutNavigationPage1": "Expand / Collapse Toggle; Check / Uncheck Checkbox",
+	"popupShortcutNavigationPage1": "Expand / Collapse Toggle",
 	"popupShortcutNavigationPage2": "Go down one line",
 	"popupShortcutNavigationPage3": "Go up one line",
 	"popupShortcutNavigationPage4": "Go to the start of the line",
@@ -1138,6 +1138,7 @@
 	"popupShortcutNavigationPage7": "Go to the end of the Object",
 	"popupShortcutNavigationPage8": "Move selected block(s) around",
 	"popupShortcutNavigationPage9": "Open relations view",
+	"popupShortcutNavigationPage10": "Check / Uncheck Checkbox",
 	
 	"popupShortcutMarkdownWhileTyping": "While typing",
 	"popupShortcutMarkdownWhileTyping1": "Inline code",
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -1129,7 +1129,7 @@
 	"popupShortcutNavigationMenu4": "Go to the right side of navigation. Link to page",
 	"popupShortcutNavigationMenu5": "Select option",
 	"popupShortcutNavigationPage": "Editor navigation",
-	"popupShortcutNavigationPage1": "Expand / Collapse Toggle",
+	"popupShortcutNavigationPage1": "Expand / Collapse Toggle; Check / Uncheck Checkbox",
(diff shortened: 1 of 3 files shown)


src/json/text.json
```diff
@@ -1129,7 +1129,7 @@
 	"popupShortcutNavigationMenu4": "Go to the right side of navigation. Link to page",
 	"popupShortcutNavigationMenu5": "Select option",
 	"popupShortcutNavigationPage": "Editor navigation",
-	"popupShortcutNavigationPage1": "Expand / Collapse Toggle",
+	"popupShortcutNavigationPage1": "Expand / Collapse Toggle; Check / Uncheck Checkbox",
```
ra3orblade, line 1132: This is not correct way to add strings, it's 2 different strings one of which is already translated into 20 languages.

src/ts/component/editor/page.tsx
```diff
@@ -793,10 +793,18 @@ const EditorPage = observer(class EditorPage extends React.Component<Props, Stat
 			});
 
 			if (block.isTextToggle()) {
-				keyboard.shortcut(`${cmd}+shift+t`, e, () => {
```
ra3orblade, line 796: Changing the shortcut itself should be more of a feature request, I can't allow this without a product review.

src/ts/component/editor/page.tsx
```diff
@@ -793,10 +793,18 @@ const EditorPage = observer(class EditorPage extends React.Component<Props, Stat
...
 				});
 			};
+
+			if (block.isTextCheckbox()) {
+				keyboard.shortcut(`${cmd}+enter`, e, () => {
+					UtilData.blockSetText(rootId, block.id, text, marks, true, () => {
+						C.BlockTextSetChecked(rootId, block.id, !block.content.checked);
+					});
+				});
+			};
```
ra3orblade, line 807: This code should be combined with the code above, its only different in Api command

ra3orblade (changes requested):

src/json/text.json
mikailcf: Ok, there aren't any kind of guides for that, so I didn't know. Could you tell me what's the correct way of adding a new string and translations for it then?

src/ts/component/editor/page.tsx
mikailcf: I thought that might be the case. I'll revert back.

src/json/text.json
ra3orblade: There is no actual guide, but if you change something it needs to be translated again and it creates less friction when it's a new line, especially with such a case when new line is added to the same place like here.

Now I think you can leave it like this, does not matter too much.

src/ts/component/editor/page.tsx
mikailcf: I've pushed new code separating the shortcuts.

src/json/text.json
mikailcf: I see. I've push new code adding a new string and leaving the old as-is. See if it helps.

ra3orblade (approved):

ra3orblade (approved):

fuksman:
@any contributor @mikailcf code

PR #807, opened 2024-07-05
add Batch lang and some lang code aliases

hellodword:
---

---

### Description

The `jsonc` supporting is not in the official docs of prismjs, but it works great with the `JSON` in anytype-ts:

````
```jsonc
{
	// comment
	"foo": "bar", // comment
	/*
		comment
	*/
	"test": 1
}
```
````
```jsonc
{
	// comment
	"foo": "bar", // comment
	/*
		comment
	*/
	"test": 1
}
```

image

The `sh` and `md` is supported by GitHub:

````
```sh
echo "foo"
```
````

```sh
echo "foo"
```
````
```md
# h1
```
````

```md
# h1
```

I didn't sort them because I want to keep the aliases nearby, correct me if I'm wrong.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


src/json/lang.ts
```diff
@@ -41,6 +44,7 @@ export default {
 		livescript: 'LiveScript',
 		lua: 'Lua',
 		markup: 'Markup',
+		md: 'Markdown',
```
ra3orblade, line 47: What for is it? It will result in double entries in the list and it won't work as there are no corresponding components in Prism. So you will have to map md value to markdown file.

Screenshot 2024-07-05 at 09 01 48
https://github.com/PrismJS/prism/tree/master/components

ra3orblade (changes requested):

src/json/lang.ts
ra3orblade: Yes, this would be much better and this way it would be possible to find the correct option by alias and not double options with the same name. You need to check if this code works for the alias in the text block:

```
let lang = fields.lang;
			let grammar = Prism.languages[lang];

			if (!grammar && (lang != 'plain')) {
				lang = J.Constant.default.codeLang;
				grammar = Prism.languages[lang];
			};
```

ra3orblade: Because I think this map `Prism.languages` is built upon the list of required lang files, to be precise. So it may not work and you need to manually map alias to the lang file in this logic.

PR #808, opened 2024-07-05
use plain as the default language code

hellodword:
---

---

### Description

For the cases that some unsupported code snippets are pasted:

````
```bar
foo
```
````

Currently, Anytype automatically treats unsupported code snippets as `javascript`, which complicates the migration process.

I believe keeping the default as plain text is better because:

1. Anytype already remembers historical choices, so this won't burden the creation of new code snippets.

2. There is no data indicating that defaulting to JavaScript better aligns with user habits.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/form/select.tsx
+++ b/src/ts/component/form/select.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import $ from 'jquery';
-import { I, S, Relation } from 'Lib';
+import { I, S, J, Relation } from 'Lib';
 import { Icon, MenuItemVertical } from 'Component';
 
 interface Props {
@@ -68,8 +68,9 @@ class Select extends React.Component<Props, State> {
 			};
 		});
 
-		if (!current.length && options.length) {
-			current.push(options[0]);
+		if (!current.length) {
+			const { codeLang } = J.Constant.default;
+			current.push({ id: codeLang, name: J.Lang.code[codeLang] });
 		};
 
 		let onClick = null;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/ts/component/form/select.tsx [hunk 1]
```diff
@@ -68,8 +68,8 @@ class Select extends React.Component<Props, State> {
 			};
 		});
 
-		if (!current.length && options.length) {
-			current.push(options[0]);
+		if (!current.length) {
+			current.push({ id: 'plain', name: J.Lang.code['plain'] });
```
ra3orblade, line 72: Should be 

```
const { codeLang } = J.Constant.default;
{ id: codeLang, name: J.Lang.code[codeLang] }
```

ra3orblade (approved):

src/ts/component/form/select.tsx
```diff
@@ -68,8 +68,9 @@ class Select extends React.Component<Props, State> {
 			};
 		});
 
-		if (!current.length && options.length) {
-			current.push(options[0]);
+		if (!current.length) {
+			const { codeLang } = J.Constant.default;
+			current.push({ id: codeLang, name: J.Lang.code[codeLang] });
```
ra3orblade, line 73: This code breaks default value of every select in the app.

PR #810, opened 2024-07-05
feat: import all supported languages and aliases from Prism

hellodword:
---

- [x] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

### Description

closes https://github.com/anyproto/anytype-ts/pull/807

Due to some networking issues, I hadn't build and test it. So I'll draft it initially.

### What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

### Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 [tech-docs](https://github.com/anyproto/tech-docs)
- [ ] 🙅 no documentation needed

PR #813, opened 2024-07-06
use prettier as default formatter and add prettier configuration file

hellodword:
---

---

### Description

See https://github.com/anyproto/anytype-ts/pull/810#discussion_r1666966270

Since I'm unable to follow the recommended format style, I think we should provide a common rule that works.

I think this PR does not break the eslint rules:

`npx eslint src/ts/lib/util/common.ts`

And I didn't find any [rule](https://prettier.io/docs/en/options.html) that supports the semicolon after each `for` statement or `if-else` statement:

```diff
                if (!electron) {
-               };
+               }


                for (const item of items) {
-               };
+               }
```

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [x] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


ra3orblade:
I'm sorry, but unfortunately code formatting PR's are not accepted.

ra3orblade:
There is no special development environment, I use regular VSCode setup without any special settings except TabNine and Copilot. The problem with automatic formatters is that they usually format the code incorrectly as for example is in your PR. You change some small line of code and your whole file gets formatted, this is unacceptable.

THe solution is to write the style guide I guess.

ra3orblade:
I have fixed linter problems.

ra3orblade:
Didn't push it: https://github.com/anyproto/anytype-ts/commit/6467981e9948bbe067f6e445669f7b0c7f4b1997

Just tried prettier one more time, it's bad - it removed empty lines where it should not touch them, it removes semicolons even if the semicolon option is set to true, it breaks imports into multiline even when printWidth is set to 150. It's just bad.

ra3orblade:
> Agree with `Bates550` said in that post:
> 
> > at least it's consistent and everyone's code will be formatted that way. It's not like, Alice formats her function parameters on multiple lines and Bob sometimes does when it's a cloudy day. With prettier they can write it however they want (regardless of the weather) and it'll just be converted to the same thing.
> 
> but I'll learn to adapt to the style.

Thanks. I agree to this too, I will find a time to write code style to make that easier, right now the focus is a bit on other things.

PR #1207, opened 2025-02-08
update the highlight effect display for the confirm popup

ShirayukiRin:
---

---

### Description
Currently, the confirm popup default will highlight the confirm button, although you have not hover your cursor on it yet, and will not remove the last highlighted button after you move your cursor away.
I think this is a bit unintuitive and could cause confusion.

https://github.com/user-attachments/assets/59624fad-7552-4c5c-b3c0-d1eba3329878

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/popup/confirm.tsx
+++ b/src/ts/component/popup/confirm.tsx
@@ -7,7 +7,7 @@ const PopupConfirm = observer(class PopupConfirm extends React.Component<I.Popup
 
 	refButtons: any = null;
 	refCheckbox: any = null;
-	n = 0;
+	n: number = null;
 
 	constructor (props: I.Popup) {
 		super(props);
@@ -16,6 +16,7 @@ const PopupConfirm = observer(class PopupConfirm extends React.Component<I.Popup
 		this.onConfirm = this.onConfirm.bind(this);
 		this.onCancel = this.onCancel.bind(this);
 		this.onMouseEnter = this.onMouseEnter.bind(this);
+		this.onMouseLeave = this.onMouseLeave.bind(this);
 		this.setHighlight = this.setHighlight.bind(this);
 	};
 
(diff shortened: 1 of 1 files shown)


ra3orblade:
Can't accept this, it works with a keyboard and the logic you have added is not relevant unfortunately. We have already discussed it with designers and decided it should work like this. The code itself can be shortened this way:

```
either 

onMouseLeave (e: any) {
    $(this.refButtons).find('.hover').removeClass('hover');
}

or 

onMouseLeave (e: any) {
   $(e.currentTarget).removeClass('hover')
}
```

Changing n to null will result in an error with the next keypress as it adds arrow derection as `this.n += dir;`

ShirayukiRin:
Hello,
Thank you for reviewing.
I have already added a check in the keypress event, so this should prevent any errors caused by null values.

```typescript
if (this.n === null) {
	this.n = 0;
	this.setHighlight();
	return;
}
```

ShirayukiRin:
According to my tests, you can still navigate using the keyboard. I haven’t removed any functionality; I’ve just set it so that no option is selected by default. You can still use the arrow keys to select an option and press Enter to confirm.


https://github.com/user-attachments/assets/f655982a-375a-4fd8-afad-0f13a1d609f5

ShirayukiRin:
> The code itself can be shortened this way:

Thank you for your valuable suggestion! I've pushed a new commit to implement it. 😊

ra3orblade:
```
if (this.n === null) {
				this.n = 0;
				this.setHighlight();
				return;
			}
```

this is not the best solution, first of you check it in arrow shortcut, while you can press enter when n is null and will still get an error.

ShirayukiRin:
```typescript
keyboard.shortcut('enter, space', e, () => {
			e.stopPropagation();
			const buttons = $(this.refButtons).find('.button');

			if (buttons[this.n]) {  // here
				$(buttons[this.n]).trigger('click');
			};
		});
```

I'm not entirely sure, but is the error coming from here? It seems like `button[this.n]` should just return `undefined` if `this.n` is `null`, so it doesn’t seem like it would cause an error?

ra3orblade:
> ```ts
> keyboard.shortcut('enter, space', e, () => {
> 			e.stopPropagation();
> 			const buttons = $(this.refButtons).find('.button');
> 
> 			if (buttons[this.n]) {  // here
> 				$(buttons[this.n]).trigger('click');
> 			};
> 		});
> ```
> 
> I'm not entirely sure, but is the error coming from here? It seems like `button[this.n]` should just return `undefined` if `this.n` is `null`, so it doesn’t seem like it would cause an error?

It will be undefined and the enter won’t work, this is not the desired behaviour. You should be able to just press enter without hovering the button with a mouse.

PR #1221, opened 2025-02-13
Drag to select box display issues

ShirayukiRin:
---

---

### Description

This commit addresses the following issues:  
1. The Drag-To-Select box was not displaying in sets and collections.  

    https://github.com/user-attachments/assets/ba0c56b2-3347-408d-a4d3-23c1c9493c9c

2. I noticed that the selection box only appears when selecting at least two nodes. Was this intentional? Additionally, when holding the Ctrl key, the box appears immediately even with just one selection.  

    https://github.com/user-attachments/assets/be82ae22-676d-42b9-8fe9-2e0dd19a597c

I'm not entirely sure about the design of the `checkNodes` function, so this commit is just a test. If you think keeping the Drag-To-Select box always visible would be a better approach, please let me know how I should adjust the implementation.  

I've attached a video showing the behavior after my changes—does this feel more intuitive to you? (At least, I think so! 😄)

https://github.com/user-attachments/assets/e52cbbe0-797f-4b3f-8866-d36cd6481a9d

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/selection/provider.tsx
+++ b/src/ts/component/selection/provider.tsx
@@ -388,9 +388,11 @@ const SelectionProvider = observer(forwardRef<SelectionRefProps, Props>((props, 
 		};
 
 		const length = (list[I.SelectType.Block] || []).length;
-		const param = U.Router.getParam(U.Router.getRoute());
+		const isPopup = keyboard.isPopup();
+		const popupMatch = keyboard.getPopupMatch();
+		const isSet = !isPopup ? keyboard.isMainSet() : [ 'set' ].includes(popupMatch.params.action);
 
-		if ((target.current.length === 0 && !allowRect.current) || (param.page === 'main' && param.action === 'set')){
+		if ((target.current.length === 0 && !allowRect.current) || (isSet)){
 			allowRect.current = true;
 		};
 
--- a/src/ts/component/selection/provider.tsx
+++ b/src/ts/component/selection/provider.tsx
@@ -388,8 +388,9 @@ const SelectionProvider = observer(forwardRef<SelectionRefProps, Props>((props, 
 		};
 
 		const length = (list[I.SelectType.Block] || []).length;
+		const param = U.Router.getParam(U.Router.getRoute());
 
-		if ((target.current.length === 0 && !allowRect.current) || (target.current[0] && target.current[0].id.includes('dataview'))){
+		if ((target.current.length === 0 && !allowRect.current) || (param.page === 'main' && param.action === 'set')){
 			allowRect.current = true;
 		};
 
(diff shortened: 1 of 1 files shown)


ra3orblade:
No, the idea of the flag was to hide the box when selecting text, but I forgot the case of set and collections. This fix is not good, since it will show it all the time and it's annoying.

ShirayukiRin:
Is this the situation with text selection? If it were to always be displayed, it could indeed be quite bothersome. 

https://github.com/user-attachments/assets/ddef4eb4-179d-4827-8b90-5e27ca589859



What do you think about changing it so that it only appears when selected from outside?

ShirayukiRin:
How does this sound? 
When selecting from the text box, at least two blocks need to be selected for the drag-to-select box to appear. 
However, if you start dragging from the outside edges of the blocks, it will display immediately. This also applies to sets and collections.


https://github.com/user-attachments/assets/3bf408ef-4f9e-4459-b545-cfd96f110071

ra3orblade:
> How does this sound? When selecting from the text box, at least two blocks need to be selected for the drag-to-select box to appear. However, if you start dragging from the outside edges of the blocks, it will display immediately. This also applies to sets and collections.
> 
>  video.mp4

this looks fine from my perpective

ShirayukiRin:
I've made the changes. Does this look correct to you?

ra3orblade:
> I've made the changes. Does this look correct to you?

No, its not correct, selection can work in the popup and behaviour should be different, logic should look something like this:

```
 const isPopup = keyboard.isPopup();
 const popupMatch = keyboard.getPopupMatch();
 const isSet = !isPopup ?  keyboard.isMainSet() : [ 'set' ].includes(popupMatch.params.action);
``

**Comment by ShirayukiRin** [comment]
Updated! 😊

File: src/ts/component/selection/provider.tsx [hunk 1]
```diff
@@ -387,6 +388,13 @@ const SelectionProvider = observer(forwardRef<SelectionRefProps, Props>((props,
 		};
 
 		const length = (list[I.SelectType.Block] || []).length;
+		const isPopup = keyboard.isPopup();
+		const popupMatch = keyboard.getPopupMatch();
+		const isSet = !isPopup ? keyboard.isMainSet() : [ 'set' ].includes(popupMatch.params.action);
+
+		if ((target.current.length === 0 && !allowRect.current) || (isSet)){
```
ra3orblade, line 395: - Remove parenthesis from `(isSet)`, they are redundant
- No need to check length === 0, you can just check `!target.current.length`

ra3orblade (changes requested):

src/ts/component/selection/provider.tsx
ShirayukiRin: Done! 😊

ra3orblade (approved):

src/ts/component/selection/provider.tsx
ra3orblade: Thanks, good job, works good as I see

PR #1606, opened 2025-08-15
Move files to bin from blocks2

Shampra:
---

---

### Description
Add an action to delete file(s) directly from an object.
This action is added to the block menu and : 
- send file(s) to bin (the same action than from this file as object)
- it display the same toast window indicating the sending of file(s) to bin
- remove block

This allows files to be deleted directly from an object without having to go into the file as an object.
Deletion management (and its current drawbacks) remains strictly the same :  this just makes the deletion process easier and improves the user experience.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://community.anytype.io/t/distinct-delete-and-remove-for-files-and-relations/21576

Icon is the same as "Delete block", I think it would be nice to have a different one.
By default, I let the design team choose it, but I can make suggestions.

Multiples files :
image
image

An unique file :
image

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -102,6 +102,7 @@
 	"commonRestore": "Restore",
 	"commonRestoreFromBin": "Restore from Bin",
 	"commonMoveToBin": "Move to Bin",
+	"commonMoveNFilesToBin": "Move %s %s to Bin",
 	"commonMoveTo": "Move to",
 	"commonLinkTo": "Add Link to Object",
 	"commonUpload": "Upload",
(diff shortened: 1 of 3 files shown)


ra3orblade:
This action was added and the postponed by founders, they want it to work another way. And in your PR you have added separate case with whole logic for ObjectListArchive, but there is an action for that, Action.archive which implements all the logic.

Issue #1542, opened 2025-07-25
Chinese IME input before formatted content leads to style change.


yudi-corgi:
### Have you read a contributing guide?

- [x] I have read CONTRIBUTING.md
- [x] I have searched the existing issues and didn't find any that were similar
- [x] I have considered creating a pull request with fixes instead of a bug report and want to proceed

### Current Behavior

Hello AnyType Team, 

I've observed an issue with AnyType's rich text formatting capabilities. When new content is inserted before already formatted text (e.g., bold, italic, code), the existing formatting unexpectedly changes.

In my testing, I've noted that this behavior differs based on the input method:

- Using an English IME: Existing formatted content remains unaffected.
- Using a Chinese IME: Existing formatted content is affected and exhibits issues.

Here's a GIF demonstrating the behavior:
(The bold styling shown in the GIF is applied via custom CSS)
![Image](https://github.com/user-attachments/assets/79a44ddc-71ae-4f33-9a49-50ea8c948283)

### Expected Behavior

Regardless of where or when new content is added to a block, and irrespective of the Input Method Editor used (including Chinese IMEs), existing formatted text should retain its applied formatting without any changes. Inserting text should not affect the rendering of surrounding, already-formatted elements.

### Steps To Reproduce

1. Type some text and apply bold formatting to a portion of it.
2. Place the cursor before the bolded content, then insert new text using a Chinese IME.

### Environment

```markdown
- OS: win11 24H2
- Version: 0.47.6
```

PR #1633, opened 2025-08-26
fix: Chinese IME input before formatted content leads to style change

twinklestar03:
---

---

### Description

This PR fixes https://github.com/anyproto/anytype-ts/issues/1542, by removing the `this.setValue` in `onCompositionEnd` event under `/src/ts/component/block/text.tsx` > BlockText.
I've tested, removing `this.setValue` in `onCompositionEnd` can be a workaround fix, which not breaking normal text editing behavior. 

The bug is causing by wrong html is generated when `this.setValue()` is called from `onCompositionEnd`.
`[start, end)` stored in `this.marks` is not being updated after new characters are composited:
```js
	setValue (v: string, restoreRange?: I.TextRange) {
		// Only apply unicode replacements if not composing IME
		if (block.isTextCode()) {
			...
		} else {
			if (!keyboard.isComposition && !ignored) {
				const parsed = Mark.fromUnicode(html, this.marks, false);

				html = parsed.text;
				this.marks = parsed.marks;
			};

			html = Mark.toHtml(html, this.marks);  // Wrong html is generated
		};

		html = html.replace(/\n/g, '<br/>');
		if (this.refEditable) {
			this.refEditable.setValue(html);    // This line set wrong html

			// Restore cursor position if provided
			if (restoreRange) {
				this.refEditable.setRange(restoreRange);
			};
		};
                ...
```

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://github.com/anyproto/anytype-ts/issues/1542

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/block/text.tsx
+++ b/src/ts/component/block/text.tsx
@@ -1255,11 +1255,15 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 			this.refEditable.placeholderHide();
 		};
 	};
-	
+
 	onCompositionEnd = (e: any, value: string, range: I.TextRange) => {
 		// Use provided value and range if available, fallback to current
 		const v = value !== undefined ? value : this.getValue();
 		const r = range !== undefined ? range : this.getRange();
+		
+		// Populate marks before setValue to prevent formatting issue
+		this.marks = this.getMarksFromHtml().marks;
+		this.setValue(v, r);
 	};
 
 	onBeforeInput = (e: any) => {
--- a/src/ts/component/block/text.tsx
+++ b/src/ts/component/block/text.tsx
@@ -1260,8 +1260,6 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 		// Use provided value and range if available, fallback to current
 		const v = value !== undefined ? value : this.getValue();
 		const r = range !== undefined ? range : this.getRange();
-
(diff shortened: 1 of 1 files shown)


twinklestar03:

ra3orblade:
This line was fixing some other bug so just removing it will cause this bug it's fixing to happen again :) Marks can be updated, so maybe just fix incorrect marks ranges?

twinklestar03:
Hi, thanks for clarification. I'll propose another fix.

twinklestar03:
Hi, @ra3orblade may I ask for the details of the other bug fixed with original implementation?
I'm thinking about using a flag to indicate a composition event and only disable `setValue` when the flag is set.

twinklestar03:
I just commit a new fix which keep the `this.setValue` while maintaining the format

ra3orblade (approved):

ra3orblade:
> I just commit a new fix which keep the `this.setValue` while maintaining the format

Yeah, nice one, thanks

PR #1724, opened 2025-10-16
Specify alpha value of the background colour

BCSharp:
Replaces the RGB format of the window background colour with RGBA. The alpha channel is set to be fully opaque. This fixes some rendering issues on Gnome/Wayland.

---

---

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

[Rendering artifacts of app window shadow on Ubuntu/Wayland](https://community.anytype.io/t/rendering-artifacts-of-app-window-shadow-on-ubuntu-wayland/28998)

This problem was desktop-only. For the screen recording see the [bug report on the community forum](https://community.anytype.io/t/rendering-artifacts-of-app-window-shadow-on-ubuntu-wayland/28998).

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/electron/js/util.js
+++ b/electron/js/util.js
@@ -67,9 +67,13 @@ class Util {
 	getBgColor (theme) {
 		theme = String(theme || '');
 
+		if (is.linux && !!process.env.XDG_SESSION_TYPE && process.env.XDG_SESSION_TYPE.toLowerCase() === 'wayland') {
+			return '#00000000'; // ARGB, transparent on Wayland
+		};
+
 		const bg = {
-			'': '#f2f2f2ff',
-			dark: '#232323ff',
+			'': '#f2f2f2',
+			dark: '#232323',
 		};
 		return bg[theme];
 	};


BCSharp:

ra3orblade:
I do not get this fix, if the color is set to fully opaque then there is no reason to pass different colors at all, it can be set to one transparent color. Second thing is that if it fixes only some Linux problems it should be applied only for Linux  windows params.

BCSharp:
First, my apologies, the fix I submitted is not correct. Without checking the documentation I automatically assumed that the color format is RGBA, while the correct format is ARGB, according to the [Electron documentation](https://www.electronjs.org/docs/latest/api/browser-window#winsetbackgroundcolorbackgroundcolor).  I have visually tested it more thoroughly (Linux, macOS) and indeed it is ARGB. So the default color should not be `'#f2f2f2ff'` but `'#fff2f2f2'`. When I visually tested my first fix the difference between `f2` and `ff` was too small to notice, especially that is visible only for a split second. I am going to commit the correct version tonight (PST).

@ra3orblade As for your concerns:
> I do not get this fix, if the color is set to fully opaque then there is no reason to pass different colors at all, it can be set to one transparent color. 

We get this bug when the color does not have the alpha value specified. The color used is treated as fully opaque (as it should), but it causes a spurious frame being drawn around and outside the main window. The value of the color matters, though. It is more visible when a more noticeable color is used, say, red (`'#ff0000'`). The sequence of events when starting Anytype is then as follows: A window appears with red client area. A fraction of a second later the splash screen appears (not red but off-white, with name "anytype" and the version number). A fraction of a second later after that, a red frame appears around the main window. The red frame disappears when the title bar of the window is grabbed by the mouse.

The only way I know to avoid the spurious frame being drawn is to specify the alpha. So, continuing the example above, `'#ff0000'` becomes `'#ffff0000'`, which is the same color with the same opaqueness written in ARGB. The sequence of events is then: A window appears with red client area. A fraction of a second later the splash screen appears (with name "anytype" and version number). There is no spurious frame. This trick works regardless what the value of the alpha is, as long as it is explicitly defined. So, yes, technically speaking, you could use a fully transparent color (in which case its value is irrelevant, could be always `'#00000000'`) and the spurious frame problem goes away. But a transparent or semi-transparent color creates other problems. One is that the window appears with the client area transparent, although the window frame and decorations are opaque. Only until the splash screen is drawn, the window fills in. So for a while upon startup there is this ghost frame on the screen. It doesn't look good. The second is that, on a few occasions, just before the splash screen was drawn, I saw for a moment an Electron welcome page appear in the void transparent area before being overwritten by the splash screen. Lots of flashing and looked ugly too.

So I think it is good to keep using fully opaque colors. But to avoid window flashing, the initial background color used should be close to the color used subsequently to draw the splash screen. This is how it is implemented now, and I think it should stay this way.

> Second thing is that if it fixes only some Linux problems it should be applied only for Linux windows params.

This is not a "Linux" problem. It is hard to identify where the root of it is so that the fix will be only applied when it is strictly needed. I tested Anytype on several systems with the following results: the bug exists on Gnome/Wayland-native and Gnome/Xwayland. I could run the same system with Gnome/Xorg and the bug is not triggered, so the fix is not needed on Xorg. It looks like it is Wayland related, but some people on the forum reported no bug on KDE so maybe it is Electron/Gnome/Wayland combo only. Looking at the release notes for Electron, in v36 they switched from GTK3 to GTK4, so maybe the problem is how they interface with GTK4. But this is not Electron code, Electron inherited it from Chromium, which made the switch. It is possible to force downgrade to GTK3 using `--gtk-version=3`, but I don't know if this switch still works in Electron 38. In any case, it is ineffective on my system and does not make the bug go away. 

So here we have several large and complex open source systems: Electron, Chromium, GTK, Wayland, each of them potentially harboring the bug. I am not interested to figure out what the exact cause of it is and where to fix it for good; this is way beyond my time availability and interest. I just want to make Anytype work. I am happy that I have found a workaround. The reason why I added the alpha for all systems (rather than just Linux/Gnome/Wayland systems) is because it is much simpler this way, and harmless on other systems. ARGB is officially a valid value for setting the background of `BrowserWindow` and it works on all systems I tried. This included macOS, but excludes Windows. I don't have a proper build environment set up on Windows, and my experience of getting Anytype to build on Linux and maOS was not smooth sailing, but that's another matter. I am getting notifications from GitHub about automatic nightly builds of my PR branch, which I could use to get a Windows version built and the test it, but I would need a help from someone to get that working. I can see that you are very busy, but if you are available, I'd welcome it. Otherwise, how about merging this PR and releasing another nightly from `develop` with the fix? In the unlikely event it does not work on Windows, it's easy to revert it.

BCSharp:
OK, to avoid delay, I implemented your suggestion to use one transparent color but only on Wayland.

ra3orblade (approved):

ra3orblade:
> how about merging this PR and releasing another nightly from `develop` with the fix

sure, done

PR #1727, opened 2025-10-18
run husky.sh conditionally

BCSharp:
This is a developer's convenience change.

File `./.husky/_/husky.sh` is not in the repository so it does not exist in a clean working tree. It can be added locally (empty file is sufficient) but then it marks the working tree as dirty. It can be git-ignored, but then it is deleted when the working tree is cleaned with `git clean -fdx` and needs to be added again.

If it is important to check that the file exists (probably containing some working tree-specific checks), I can update the script that it prints a warning message.

---

---

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [x] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,8 @@
 #!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
+huskydir=$(dirname -- "$0")
+if [ -f "$huskydir/_/husky.sh" ]; then
+  . "$huskydir/_/husky.sh"
+fi
 
 npx lint-staged --concurrent false
 # npm run typecheck


ra3orblade:
I did not get the point. It is installed with `npm run prepare` and is already ignored.

BCSharp:
> It is installed with `npm run prepare` and is already ignored.

You are right. I didn't know I needed to run `npm run prepare`. On macOS, when I run `npm run dist:mac`, `prepare` is run automatically. On Linux I can't run `npm run dist:linux` due to unmet dependencies so I just build `AppImage`.

Issue #832, opened 2024-07-11
[Feature Request] Touchpad gesture navigation


larsb24:
### Have you read a contributing guide?

- [X] I have read CONTRIBUTING.md
- [X] I have searched the existing requests and didn't find any that were similar
- [X] I have considered creating a pull request instead and want to proceed

### Clear and concise description of the problem

As someone who uses browsers a lot, I have really gotten used to the two-finger swipe to navigate back.
I keep catching myself trying to move backwards to a previous document in Anytype as well but such a navigation doesn't exist unfortunately.

### Suggested solution

The ideal solution would be somehow activating the chromium two finger swipe to go backwards as the desktop app seems to be based on electron. 
I'm not sure however whether electron actually supports that feature or if it's stripped out.

### Alternative

Otherwise I suppose it would still be possible to implement such a feature with JavaScript/Typescript.

### Additional context

A forward gesture would obviously be included in that idea as well.

maksimuimin:
+1, would love to see this

artemisresende:
+1, it would improve a lot the UX

PR #1739, opened 2025-10-28
Fix back/forward mouse button press handling for Logitech mice on MacOS

scupit:
---

---

### Description

**This PR fixes an issue where the Forward and Back mouse buttons were completely ignored on MacOS when using a mouse configured with Logi Options (+).** This happens because Logitech software on MacOS maps Forward/Back to the legacy three-finger left/right swipe gestures instead of a mouse event. To reiterate, when Logi Options is installed, pressing Forward/Back on a mouse will NOT fire a mouse event; It will only fire a swipe gesture event. Since a mouse event is never fired, the existing `onMouseDown()` handler (which checks for specific button codes) is rendered useless.

This fix is MacOS-specific. Forward/Back already fire proper the proper events on Windows as expected.

There are weirdly few references to this in the wild:
- <https://news.ycombinator.com/item?id=42457468>
- See the ["Technical Notes" section here](https://sensible-side-buttons.archagon.net/) - looks like someone made a tool to solve the same system-wide years ago.

Electron has a [MacOS-specific handler for these swipe events](https://www.electronjs.org/docs/latest/api/browser-window#event-swipe-macos), so it won't conflict with event handlers on other systems. Also, since Forward/Back presses don't fire a mouse event, the `onBack()` and `onForward()` navigation functions will never be double-fired.

This had to be handled [at the BrowserWindow level](https://www.electronjs.org/docs/latest/api/browser-window#event-swipe-macos) because the swipe gesture event is never (as far as I can tell) fired as a DOM event or a [webContent "before-mouse-event"](https://www.electronjs.org/docs/latest/api/web-contents#event-before-mouse-event).

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

Just to be clear, this does NOT solve <https://github.com/anyproto/anytype-ts/issues/832>. That issue would require more granular handling of the modern two-finger touch events instead of the "legacy" swipe event this solution had to use.

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

None

### DIFF
--- a/electron/js/window.js
+++ b/electron/js/window.js
@@ -67,6 +67,22 @@ class WindowManager {
 		win.on('enter-full-screen', () => Util.send(win, 'enter-full-screen'));
 		win.on('leave-full-screen', () => Util.send(win, 'leave-full-screen'));
 
+    // Logitech software binds dedicated forward/back mouse buttons to the legacy
+    // three-finger swipe gestures on MacOS by default. The system will NOT fire mouse events
+    // for forward/back mouse button presses when this is the case. The electron swipe event is
+    // MacOS-specific, and won't interfere with Windows/Linux event handling.
+    // https://www.electronjs.org/docs/latest/api/browser-window#event-swipe-macos
+    win.on('swipe', (_e, direction) => {
+      switch (direction) {
+        case 'left':
+          Util.send(win, 'macosMouseNavigation', 'back');
+          break;
+        case 'right':
+          Util.send(win, 'macosMouseNavigation', 'forward');
+          break;
+      }
+    });
+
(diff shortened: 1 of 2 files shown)


scupit:

ra3orblade (approved):

ra3orblade:
Hello, thanks, I've merged it and refactored a little bit: https://github.com/anyproto/anytype-ts/commit/5b75b05da63195ea2dda0ebf29fa4186c8b90142

Issue #1761, opened 2025-11-16
Pasting in code block causes unexpected scroll to top


mottcha:
### Have you read a contributing guide?

- [x] I have read CONTRIBUTING.md
- [x] I have searched the existing issues and didn't find any that were similar
- [x] I have considered creating a pull request with fixes instead of a bug report and want to proceed

### Current Behavior

When pasting text into a large, multi-line code block, the editor's scroll position unexpectedly jumps to the very top of that block.

After the paste action, the editor attempts to scroll to the pasted content, but the scroll position seems to be incorrectly anchored to the beginning of the block. As a result, any paste operation within a long code block (which is handled as a single block internally) forces the viewport to jump to the top, away from the area being edited.

This behavior also occurs in callout blocks.

### Expected Behavior

After pasting text within a code block, the viewport should remain focused on the location of the pasted content. The scroll position should not jump to the beginning of the block.

### Steps To Reproduce

1. Create a new code block.
2. Paste a large block of text into it (e.g., more than a screenful of code).
3. Scroll down to the middle of the code block.
4. Place the cursor somewhere in the middle and paste another line of text.
5. Observe that the editor's scroll position immediately jumps to the first line of the code block, losing your focus.

https://github.com/user-attachments/assets/fccbcd89-e4a5-424a-9229-4533bf0b4e8d

### Environment

```markdown
- OS: Windows 11 / Ubuntu 24.04
- Version: 0.50.59-alpha

I'm debugging on Windows 11 by running the Linux version of the app in WSL2 with hot-reloading via WSLg.﻿
but the bug also reproduces on standard Windows and Linux environments.﻿
```

### Anything else?

This issue appears to be related to how focus is handled after a paste event. While normal text paragraphs don't pose a significant problem (as they are often treated as separate blocks per line), the single-block nature of code and callout blocks makes this bug particularly disruptive.

**Suspected Cause:**
I suspect the issue may be directly related to `this.focus()` being called within the `onPaste` function, located around line 1996 in `/src/ts/component/editor/page.tsx`.

https://github.com/anyproto/anytype-ts/blob/e8a7e38417cc4520171c766a9d3b9507c19e198b/src/ts/component/editor/page.tsx#L1991-L1998

**Debugging Attempt:**
[...]

PR #1767, opened 2025-11-18
Prevent unnecessary scroll in large blocks

sleroq:
---

---

### Description

This is fix for #1761 ([forum link](https://community.anytype.io/t/ui-scroll-position-jumps-when-pasting-into-long-code-blocks/27658))

`focus.scroll()` is called on every paste, which seems reasonable, so I've tried to improve the scroll implementation.

However, it's very possible that somewhere in the app we call `focus.scroll()` expecting it to move us to the top of the block even if it's already visible. I haven't found any usages like this though.

An alternative solution is probably to check the number of inserted blocks on `onPaste()`, and if it's equal to 1, we don't need to scroll. But then we need to fix undo behavior separately.

off topic:
The most annoying issue with the current focus implementation is that it still can only scrolls to the top of the block instead of the caret position. Because of this pasting large blocks of text results in losing the caret somewhere below the viewport. 

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/lib/focus.ts
+++ b/src/ts/lib/focus.ts
@@ -124,11 +124,20 @@ class Focus {
 		const container = U.Common.getScrollContainer(isPopup);
 		const ch = container.height();
 		const no = node.offset().top;
+		const nh = node.outerHeight();
 		const hh = J.Size.header;
 		const o = J.Size.lastBlock + hh;
 		const st = container.scrollTop();
 		const y = no - container.offset().top + st;
-
+		
+		const topViewport = st + hh;
+		const bottomViewport = st + ch - o;
+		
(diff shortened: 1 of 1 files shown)


ra3orblade:
The proper solution should be to change focus.scroll to work with selectionRect coordinates instead of focusable block element, this is acceptable solution for now but I have a task to rework the method to work like you have described.

PR #1775, opened 2025-11-20
Fix displaying empty relations

sleroq:
---

---

### Description

Fix for rendering empty selects which remain after object is deleted.

How to reproduce the bug with empty relations:
1. Add an object to the select on multiple notes
2. Delete the object completely
3. It will disappear only from the note you were in when deleting the object

At first seems like a backend bug, relation is not removed from all other objects. But looking at backend looks like it's intentional so just hiding the select then.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

<details>
<summary>video on how to reproduce the bug</summary>

https://github.com/user-attachments/assets/92bcbb67-dce5-4672-8cb2-64d52431545f

</details>

Screenshot 2025-11-21 at 1 33 44 AM
Screenshot 2025-11-21 at 1 33 59 AM

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/cell/index.tsx
+++ b/src/ts/component/cell/index.tsx
@@ -45,7 +45,10 @@ const Cell = observer(forwardRef<I.CellRef, Props>((props, ref) => {
 		};
 
 		if ([ I.RelationType.Select, I.RelationType.MultiSelect ].includes(relation.format)) {
-			return !!Relation.getOptions(record[relation.relationKey]).length;
+			const options = Relation.getOptions(record[relation.relationKey])
+				.filter(it => !it._empty_ && !it.isArchived && !it.isDeleted);
+			
+			return !!options.length;
 		};
 
 		return Relation.checkRelationValue(relation, record[relation.relationKey]);


src/ts/component/cell/index.tsx
```diff
@@ -39,7 +39,19 @@ const Cell = observer(forwardRef<I.CellRef, Props>((props, ref) => {
...
 		};
-		return isName ? true : Relation.checkRelationValue(relation, record[relation.relationKey]);
+
+		if (isName) {
+			return true;
+		};
+
+		if ([ I.RelationType.Select, I.RelationType.MultiSelect ].includes(relation.format)) {
+			const options = Relation.getOptions(record[relation.relationKey])
+				.filter(it => !it._empty_ && !it.isArchived && !it.isDeleted);
```
sleroq: Filter part is not required (array is empty), but it's used in other places so I assume it's useful.

Issue #1814, opened 2025-12-07
Collapsible Headers (Notion/Google Docs-style)


ugosan:
### Have you read a contributing guide?

- [x] I have read CONTRIBUTING.md
- [x] I have searched the existing requests and didn't find any that were similar
- [x] I have considered creating a pull request instead and want to proceed

### Clear and concise description of the problem

I have implemented Collapsible Headers as described [here](https://community.anytype.io/t/toggles-for-headers-collapsible/1318)

**Behavior**
H1 collapses everything until the next H1
H2 collapses everything until the next H2 or H1
H3 collapses everything until the next H3, H2, or H1
Arrow shows at end of header line, hidden when expanded (visible on hover), always visible when collapsed
State is persisted, defaults to expanded

**Benefits**
Better navigation in long documents
Familiar UX from Notion and Google Docs
Drastically reduces visual clutter

### Suggested solution

I am creating collapsible headers (H1, H2, H3) that work like Notion and Google Docs.

ra3orblade:
This is implemented in latest build differently and works the same way it works in Notion.

PR #1815, opened 2025-12-07
Collapsible Headers

ugosan:
---

---

### Description

This PR implements hierarchical collapsible headers (H1, H2, H3) with visual chevron indicators and persisted state.

Users can now organize long documents by collapsing header sections, reducing visual clutter while maintaining full content access. The hover-based arrow for expanded headers keeps the interface clean while providing clear affordance for the collapse action.

All existing headers default to expanded state, ensuring no disruption to existing documents.

### Features

**Hierarchical Collapsing**
- **H1** collapses all content blocks until the next H1
- **H2** collapses all content until the next H2 or H1
- **H3** collapses all content until the next H3, H2, or H1

**Visual Indicators**
- Chevron arrow displayed at the right end of header lines
- **Collapsed state**: Arrow always visible, pointing right (→)
- **Expanded state**: Arrow hidden by default, appears on hover pointing down (↓)
- Smooth rotation animation between states using CSS transitions

**State Persistence**
- Collapse/expand state persisted to local storage
- Pre-existing headers default to expanded state on first load
- State maintained across page reloads and sessions

**Selection & Deletion**
- Selecting a collapsed header automatically includes all section content
- Deleting a collapsed header removes the header and all hidden blocks in its section
- Works with drag menu, context menu, and keyboard shortcuts

**Files Modified:**
- `src/ts/store/block.ts`: Added `toggleHeader()`, `getHeaderSectionBlocks()`, and `isInsideCollapsedHeader()` methods
- `src/ts/component/block/index.tsx`: Integrated header toggle initialization and arrow rendering
- `src/scss/component/editor.scss`: Added chevron arrow styling with hover states
- `src/ts/component/selection/provider.tsx`: Modified `getForClick()` to include section blocks for collapsed headers
- `src/ts/component/editor/page.tsx`: Updated `blockRemove()` to delete section content with collapsed headers

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://github.com/anyproto/anytype-ts/issues/1814
https://community.anytype.io/t/toggles-for-headers-collapsible/1318

https://github.com/user-attachments/assets/353fb414-1328-46d3-a28b-fd0b86dbc0c1

**Arrow is displayed upon mouse hover to signal it can be collapsed, invisible otherwise**
Screenshot 2025-12-07 at 10 03 28 AM

**Arrow is always visible if header is collapsed, to signal there is content inside**
Screenshot 2025-12-07 at 10 03 34 AM

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/component/editor/page.tsx
+++ b/src/ts/component/editor/page.tsx
@@ -1225,7 +1225,7 @@ const EditorPage = observer(class EditorPage extends React.Component<Props, Stat
 
 		const { rootId } = this.props;
 		const dir = pressed.match(Key.up) ? -1 : 1;
-		const next = S.Block.getFirstBlock(rootId, -dir, it => it.isFocusable());
+		const next = S.Block.getFirstBlock(rootId, -dir, it => it.isFocusable() && !S.Block.isBlockHidden(rootId, it.id));
 
 		this.focusNextBlock(next, dir);
 	};
@@ -1248,7 +1248,7 @@ const EditorPage = observer(class EditorPage extends React.Component<Props, Stat
 			ids[method]();
 		} else {
 			const idx = (dir < 0) ? 0 : idsWithChildren.length - 1;
-			const next = S.Block.getNextBlock(rootId, idsWithChildren[idx], dir, it => !it.isSystem());
+			const next = S.Block.getNextBlock(rootId, idsWithChildren[idx], dir, it => !it.isSystem() && !S.Block.isBlockHidden(rootId, it.id));
 
 			method = dir < 0 ? 'unshift' : 'push';
 			if (next) {
@@ -1568,9 +1568,9 @@ const EditorPage = observer(class EditorPage extends React.Component<Props, Stat
 			if (!next) {
 				// If block is closed toggle - find next block on the same level
 				if (block && block.isTextToggle() && !Storage.checkToggle(rootId, block.id)) {
-					next = S.Block.getNextBlock(rootId, focused, dir, it => (it.parentId != block.id) && it.isFocusable());
+					next = S.Block.getNextBlock(rootId, focused, dir, it => (it.parentId != block.id) && it.isFocusable() && !S.Block.isBlockHidden(rootId, it.id));
 				} else {
-					next = S.Block.getNextBlock(rootId, focused, dir, it => it.isFocusable());
+					next = S.Block.getNextBlock(rootId, focused, dir, it => it.isFocusable() && !S.Block.isBlockHidden(rootId, it.id));
 				};
 			};
 
@@ -1642,7 +1642,7 @@ const EditorPage = observer(class EditorPage extends React.Component<Props, Stat
 			} else {
 				const nextIdx = dir > 0 ? rowElement.childrenIds.length - 1 : 0;
 
-				next = S.Block.getNextBlock(rootId, rowElement.childrenIds[nextIdx], dir, it => it.isFocusable());
+				next = S.Block.getNextBlock(rootId, rowElement.childrenIds[nextIdx], dir, it => it.isFocusable() && !S.Block.isBlockHidden(rootId, it.id));
 				cb();
 			};
 		} else {
@@ -2156,7 +2156,7 @@ const EditorPage = observer(class EditorPage extends React.Component<Props, Stat
 	
 	blockMerge (focused: I.Block, dir: number, length: number) {
 		const { rootId } = this.props;
-		const next = S.Block.getNextBlock(rootId, focused.id, dir, it => it.isFocusable());
+		const next = S.Block.getNextBlock(rootId, focused.id, dir, it => it.isFocusable() && !S.Block.isBlockHidden(rootId, it.id));
 		if (!next) {
 			return;
 		};
@@ -2200,7 +2200,7 @@ const EditorPage = observer(class EditorPage extends React.Component<Props, Stat
 				};
 
 				if (dir < 0) {
-					const next = S.Block.getNextBlock(rootId, focused.id, -1, it => it.isFocusable());
+					const next = S.Block.getNextBlock(rootId, focused.id, -1, it => it.isFocusable() && !S.Block.isBlockHidden(rootId, it.id));
 					if (next) {
 						const nl = dir < 0 ? next.getLength() : 0;
 						this.focus(next.id, nl, nl, false);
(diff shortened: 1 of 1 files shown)


ugosan:

ra3orblade:
Hello, I've checked the code, it does not take into account column layouts, so there can be header, after the header column layout and inside column layout there might be a header as well and keyboard block traversing, if the header is collapsed the code will continue to traverse blocks even if they are hidden.

src/ts/component/block/index.tsx
```diff
@@ -424,6 +424,11 @@ const Block = observer(class Block extends React.Component<Props> {
 					{object}
 					{additional ? <div className="additional">{additional}</div> : ''}
 
+					{/* Arrow indicator for headers */}
+					{block.isTextHeader() ? (
+						<div className="collapseArrow" />
```
ra3orblade: this can be set as additional or marker, no need for separate element

src/ts/store/block.ts
```diff
@@ -707,6 +707,115 @@ class BlockStore {
...
+				if (isCollapsed) {
+					const sectionBlocks = this.getHeaderSectionBlocks(rootId, block.id);
+					if (sectionBlocks.includes(blockId)) {
+						return true;
+					}
+				}
+			}
+		}
+
+		return false;
```
ra3orblade: this code does not count row/column layouts and tables

ra3orblade:
I think the logic for initializing this state should work the same way as in the numbered lists:
1. Editor opens, initialization is called in the onObjectView method in dispatcher,
2. When new event comes with block structure being changed or new blocks added / deleted the hidden blocks state should be updated, the best way is to update them in the part that changed, not the whole document.
3. Current implementation with range checking in each block is very heavy and will dramatically affect the performance.

src/ts/store/block.ts
ugosan: I tried but couldn't add a header to a table, it does work with column layout (which I had no idea it was possible! thank you)

https://github.com/user-attachments/assets/81b6db6e-e857-4e37-b6eb-a20dafd4ad3f

ra3orblade:
Screenshot 2025-12-11 at 15 12 14
you have added a file that is not required

src/ts/component/block/text.tsx
```diff
@@ -210,10 +222,11 @@ const BlockText = observer(class BlockText extends React.Component<Props> {
 					onCompositionEnd={this.onCompositionEnd}
 					onBeforeInput={this.onBeforeInput}
 				/>
+
 			</div>
 		);
-	};
-	
+	};;
```
ra3orblade: double semicolon

src/ts/store/block.ts
```diff
@@ -707,6 +716,262 @@ class BlockStore {
...
+		const sectionBlocks: string[] = [];
+		
+		for (let i = headerIndex + 1; i < allBlocks.length; i++) {
+			const block = allBlocks[i];
+			
+			if (block.isTextHeader()) {
+				const blockLevel = block.isTextHeader1() ? 1 : block.isTextHeader2() ? 2 : 3;
+				if (blockLevel <= headerLevel) {
+					break;
+				}
```
ra3orblade: Please fix formatting

src/ts/store/block.ts
```diff
@@ -707,6 +716,262 @@ class BlockStore {
...
+			sectionBlocks.forEach(blockId => hiddenSet.add(blockId));
+		});
+
+		this.collapsedHeadersMap.set(rootId, collapsedSet);
+		this.hiddenBlocksMap.set(rootId, hiddenSet);
+	};
+
+	/**
+	 * Update hidden state when blocks are added (called from dispatcher on BlockAdd event).
+	 * Only recalculates for newly added blocks.
```
ra3orblade: This logic is wrong, all checks should be implemented only on structure changes, when adding or deleting blocks the structure changes anyway. So you need to get collapsed headers and hide blocks accordingly

ra3orblade:
I've added another bunch of review comments, but overall logic is wrong. Please check updateNumbersTree in StoreBlock, the traversing of the tree should happen exactly the same way, hiding/unhiding blocks during structure changes should follow the same logic, it now works correctly for numbered blocks. No need to maintain hidden/added/deleted/changed/collapsed blockIds during changes since there are only 3 possible entry points:
1. Initialisation: we traverse the tree and init which blocks are hidden and make them hidden
2. Structure change: we traverse the tree and do the same thing, structure changes can happen in multiple ways - dragndrop, layout change when creating columns, another device and etc., this mechanism should be efficient.
3. When some header is deleted or changed to another style we can remove it from toggle list safely. This should happen inside S.Block.delete/updateContent.

Once again, please check S.Block.updateNumbers logic, it's called only twice in the dispatcher - once in onObjectView, once in the event handler when block changes happen, it's sufficient to do the same logic without overcomplicating.

ugosan:
> I think the logic for initializing this state should work the same way as in the numbered lists:
> 
>     1. Editor opens, initialization is called in the onObjectView method in dispatcher,
> 
>     2. When new event comes with block structure being changed or new blocks added / deleted the hidden blocks state should be updated, the best way is to update them in the part that changed, not the whole document.
> 
>     3. Current implementation with range checking in each block is very heavy and will dramatically affect the performance.

Thanks @ra3orblade, the way I am thinking headers must hide/show elements in between headers and elements are not nested to a header, hence the range check.

ra3orblade:
> > I think the logic for initializing this state should work the same way as in the numbered lists:
> > ```
> > 1. Editor opens, initialization is called in the onObjectView method in dispatcher,
> > 
> > 2. When new event comes with block structure being changed or new blocks added / deleted the hidden blocks state should be updated, the best way is to update them in the part that changed, not the whole document.
> > 
> > 3. Current implementation with range checking in each block is very heavy and will dramatically affect the performance.
> > ```
> 
> Thanks @ra3orblade, the way I am thinking headers must hide/show elements in between headers and elements are not nested to a header, hence the range check.

Yes, but it can be done on the editor level, not block level, this is what I mean.

ra3orblade:
Closing in favor of this implementation, it needs some updates on keyboard handling but fundamentally it works. https://github.com/anyproto/anytype-ts/pull/1876/files

Issue #83, opened 2023-07-20
Publish linux ARM binaries


IdacraBackup:
Would be great to publish binaries for Linux ARM in releases, I see in package.json there is already a target "dist:linuxarm"

developomp:
Also reported here: https://community.anytype.io/t/linux-arm-support/9234

ra3orblade:
it's not building correctly

ra3orblade:
You can try, middleware builds arm version I think and it should already download correctly download the required arch release in update.sh

Vortigern-The-Grey:
Would also be interested in this as code from a Raspberry Pi 4 running Kali Linux.

afro-systems:
I've nearly been able to build this on Asahi Linux, however ran into a snag with primordials being undefined. Nearly completed troubleshooting. When I get it to build successfully, I'll submit my fixes/writeup here.

afro-systems:
> I've nearly been able to build this on Asahi Linux, however ran into a snag with primordials being undefined. Nearly completed troubleshooting. When I get it to build successfully, I'll submit my fixes/writeup here.

Sorry, been a bit busy with other stuff. I plan to reattack this week.

Abijeet:
I built the app on Asahi Linux running on an Apple MacBook Pro (16-inch, M1 Max, 2021)

![anytype-m1-pro](https://github.com/user-attachments/assets/c851cf6e-6376-4431-b21b-202e0b2ae0e0)

Here are relevant instructions (very closely mirrors what's mentioned in the [README.md](https://github.com/anyproto/anytype-ts) file):
```shell
# Installed node using nvm: 
$ node -v
v22.6.0

# install dependencies
$ sudo dnf install libsecret jq libsecret-devel make g++

# clone the anytype-ts repo and install node dependencies
$ git clone https://github.com/anyproto/anytype-ts
$ cd anytype-ts
# checkout to the latest release...in this case, I checked out to 0.42.4
$ git checkout tags/0.42.4 -b 0.42.4
$ npm install -D

# install middleware library and protobuf bindings
$ ./update.sh ubuntu-latest arm
$ mv linux-arm/anytypeHelper dist/

# build
$ npm run update:locale
$ npm run dist:linuxarm
```

The build fails with the following error message but the `dist` folder does have an `Anytype-0.42.4-arm64.AppImage` file that can be run.

```
[...]

PR #1845, opened 2025-12-28
GO-1234 feat: limit linux build targets to AppImage to support ARM64

pjjjv:
This PR limits the default linux build targets to AppImage. This avoids build failures on ARM64 systems where fpm (required for deb/rpm) is only available as an x86 binary. Related to https://github.com/anyproto/anytype-ts/issues/83


pjjjv:
This is optional.

PR #1849, opened 2026-01-05
Genearlize `isCut` for future support of Copy As Reference

goretkin:
---

- [x] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

### Description

Change from an `isCut` bool to an enum with `Copy` and `Cut` options. This does not change anything else, but
- makes the code a bit more legible
- supports new copy modes, such as "Copy As Reference", which might be a possible interface for supporting transclusions / sync blocks.

### What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [x] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

### Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 [tech-docs](https://github.com/anyproto/tech-docs)
- [x] 🙅 no documentation needed

PR #1851, opened 2026-01-05
[WIP] Transclusions (sync blocks)

goretkin:
---

---

### Description
This PR depends on
- https://github.com/anyproto/anytype-ts/pull/1849
- https://github.com/anyproto/anytype-heart/pull/2878

[Kooha-2026-01-05-02-09-35.webm](https://github.com/user-attachments/assets/9eb6d6c6-3ddb-439e-b9d7-d47077bc9491)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -69,10 +69,23 @@ Anytype is an Electron-based desktop application with TypeScript/React frontend 
 
 ### Key Architectural Patterns
 
+**Objects:**
+- An **object** is the fundamental entity in Anytype with a unique ID
+- Objects contain blocks (their content structure) and have metadata (name, type, relations, etc.)
+- Objects have two key properties:
+  - **Layout** (fixed enum, `src/ts/interface/object.ts:ObjectLayout`) - Defines presentation/structure
+  - **Type** (user-extensible)
+    - Types are themselves objects with `layout: ObjectLayout.Type`
+    - Users can create new types; Anytype provides built-in library types
+- Object metadata stored in `S.Detail`, blocks stored in `S.Block` (both keyed by object ID)
+- When working within an object, its ID is called `rootId` (frontend) or `contextId` (middleware)
+
 **Block-Based Documents:**
-- Documents are composed of blocks (text, images, databases, etc.)
+- A document (informal term) is an object.
+- Objects contain blocks (text, images, databases, etc.) that form their content
 - Each block type has corresponding model, content, and component
 - Block operations handled via gRPC commands
+- Blocks have IDs unique within their containing object
 
 **MobX State Management:**
 - Reactive state with MobX stores
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -239,6 +239,7 @@
 	"commonSaveAsFile": "Save as file",
 	"commonDownloadCode": "Download Code",
 	"commonCopyMedia": "Copy Media",
+	"commonCopyAsReference": "Copy As Reference",
 	"commonImage": "Image",
 	"commonOptions": "Options",
 
@@ -664,6 +665,10 @@
 	"blockTextAudio": "Upload playable audio",
 	"blockTextPdf": "Upload a PDF file",
 	"blockTextBookmark": "Save your favorite link with summary",
+	"blockTransclusionLoading": "Loading block...",
+	"blockTransclusionNotFound": "Referenced block not found",
+	"blockTransclusionError": "Error loading block",
+	"blockTransclusionDepthLimit": "Nested transclusion omitted.",
 	"blockTextCode": "Capture code snippet ",
 	"blockTextLine": "Separate information with line",
 	"blockTextDot": "Separate information with three dots",
--- a/src/scss/block/common.scss
+++ b/src/scss/block/common.scss
@@ -129,6 +129,7 @@
 @import "./iconUser";
 @import "./layout";
 @import "./link";
+@import "./transclusion";
 @import "./media";
 @import "./relation";
 @import "./table";
--- a/src/scss/block/transclusion.scss
+++ b/src/scss/block/transclusion.scss
@@ -1,0 +1,64 @@
+.block.blockTransclusion {
+	.depthLimit {
+		display: flex;
+		align-items: center;
+		padding: 12px;
+		background: var(--color-shape-tertiary);
+		border-radius: 8px;
+		cursor: help;
+
+		.icon {
+			margin-right: 8px;
+			color: var(--color-text-secondary);
+		}
+
+		.name {
+			font-family: monospace;
+			color: var(--color-text-secondary);
+			font-size: 14px;
+		}
+
+		&:hover {
+			background: var(--color-shape-highlight-medium);
+		}
+	}
+
+	.loading, .notFound, .error {
+		display: flex;
+		align-items: center;
+		padding: 12px;
+
+		.icon, .loader {
+			margin-right: 8px;
+		}
+
+		.name {
+			color: var(--color-text-secondary);
+			font-size: 14px;
+		}
+	}
+
+	.transcludedContent {
+		position: relative;
+		padding-left: 8px;
+		border-left: 2px solid var(--color-control-bg);
+
+		.transclusionIndicator {
+			position: absolute;
+			left: -10px;
+			top: 8px;
+
+			.icon {
+				color: var(--color-text-secondary);
+				font-size: 12px;
+			}
+		}
+
+		// Make transcluded blocks visually distinct
+		opacity: 0.9;
+
+		&:hover {
+			opacity: 1;
+		}
+	}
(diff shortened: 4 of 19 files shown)


goretkin:
Copied from a README in the PR:


# Running notes for implementing transclusions

- the feature has been requested many times
- the canonical source of the discussion is at https://community.anytype.io/t/add-block-linking-synced-embedded-blocks/1360/8

# Potential blockers
- 2023-nov-27 Townhall
    - https://community.anytype.io/t/add-block-linking-synced-embedded-blocks/1360/37
    - cryptography makes it challenging to link to a block in another page
    - resolved by using Any Store. [GO-3663](https://github.com/anyproto/anytype-heart/pull/1374)
- 2024-jan-30 discourse
    - https://community.anytype.io/t/add-block-linking-synced-embedded-blocks/1360/48
    - The problem with this general feature request is that our current architecture does not allow editing object blocks while you are in another object. It only allows for previewing other objects and editing their relationships. Therefore, we are considering adding a new layout, ‘block or citation,’ which would consist of a single paragraph serving as the name of this object. This way, we can transform any block into an object with this layout, allowing it to be transcluded or collected in a set. You will be able to tag them, etc. However, at this stage, it will not be possible to use markup for this block
- 2024-aug-26
    - https://community.anytype.io/t/add-block-linking-synced-embedded-blocks/1360/74
    -  Yes! We just moved to a new datastore which would allow us to make transclusions easily. We start to use it with chats, than move collections to it, later it’ll be editor. We may start to work on it the beginning of the next year

# Related back-end features
## `Content.Bookmark`
- `* Bookmark is to keep a web-link and to preview a content.`
- information comes from `targetObjectId`

## `Content.Link`
- `* Link: block to link some content from an external sources.`
- `targetBlockId` is a misnomer. It is an object link.

# Justification for BlockTransclusion React component
It needs to load other objects. We might instead prefer to have a dataloading phase, and then no separate React component for transclusions.

# Larger questions
- Is there any infrastructure for deep links blocks?
    - Can that same referencing mechanism be used for transclusions?
- Does Anytype have a concept of block grouping?
- Why is there interest in "blocks as objects"? Objects can be linked to already. The UI has a feature where you can turn a block into an object. But it changes its appearance.
- How does Anytype handle beta features generally? Are there feature flags? Is there anything like "this feature is in development"? Antype 'labs'?


# Terminology note
In the vernacular of "linking", the link *target* represents the *source* of truth.

# Design
## How do you distinguish between copy-pasting as a duplicate versus pasting as a transclusion?
At the moment, it is decide-on-copy.

### Decide on paste
Prompt similar to pasting URL
- maybe each transclusion source should be an object, just like a URL bookmark is an object.

### Decide on copy
Make a new "copy as reference" command that does the same thing, but sets some state in the clipboard to indicate the desired pasting.


## Should the source block indicate that it is transcluded?
Currently, no.

## How should the system handle "recursive transclusions"
add a toggle for expanding?
persist the toggle state?
no toggle at first. only <RECURSIVE TRANSCLUSION> + a tooltip that shows "this feature of Anytype is in development".

## How to link back to source block?

# Future features
- indicate when a block is a transclusion source
- warn/error on deleting source block that has transclusion references
    - deleting a bookmark or object permanently shows `Non-existent object` with no warning, so maybe transclusions should behave the same.
- live updates in the front end
- change source block (swap a transclusion and the original, updating all transclusions)

ra3orblade:
This won't work on a large scale: 
1. ObjectShow is extremely slow, it builds the tree of the document, so if you will have 50, 100, 1000 transcluded blocks it will result in a non-opening document.
2. ObjectShow is not subscribed to changes, so in your scenario transcluded blocks would not receive updates.

Unfortunately I am not sure I will be able to merge this PR.

goretkin:
I appreciate the quick review, thank you!

So it sounds like an implementation of transclusions must not rely on
ObjectShow, then.

I have a few questions:

1. Does it mean there is a scale issue if there are 100s of bookmarks on a
page? I think that also uses ObjectShow.

2. The screen recording shows updating of the transclusions when I
tab/click out of the source block. There is no live updating. Does that
match your expectation?

3. What are your thoughts on a more suitable implementation strategy? The
migration to Anystore seems to have made the transclusion feature possible.
Are there other big changes necessary to enable this oft-requested feature?

4. I came across discussions about "blocks as objects" and how that might
be an interesting approach for transclusions. Is that still relevant?

ra3orblade:
> I appreciate the quick review, thank you! So it sounds like an implementation of transclusions must not rely on ObjectShow, then. I have a few questions: 1. Does it mean there is a scale issue if there are 100s of bookmarks on a page? I think that also uses ObjectShow. 2. The screen recording shows updating of the transclusions when I tab/click out of the source block. There is no live updating. Does that match your expectation? 3. What are your thoughts on a more suitable implementation strategy? The migration to Anystore seems to have made the transclusion feature possible. Are there other big changes necessary to enable this oft-requested feature? 4. I came across discussions about "blocks as objects" and how that might be an interesting approach for transclusions. Is that still relevant?

1. Bookmarks, links, dataviews and files work based on dependency subscriptions and object store, apart from dataview blocks all of them subscribe to property value changes which are indexed in the local database which then are dispatched as events for the client basically.
2. From my perspective transclusions without live updates do not make much sence, but it's more of a product question.
3. The migration to any-store happened on the database level, editor is still based on the CRDT trees and we are in the early stage of discussion on how to move it to any-store and this is a very complicated task from every point of view. 

Implementation using ObjectShow while technically possible will result in huge requests. First of all ObjectShow loads ALL blocks and ALL dependencies with all properties and creates maps in stores for them, so when you have multiple blocks transcluded you will not only build trees for these document which can be very slow at some points (imagine document for 20k blocks for example) but also create structure for every document in MobX state, which should be cleared the same way that the ordinary state is cleared when object is closed. The idea to switch to any-store was from the fact that any-store works faster and allows client in theory to work with blocks making queries like SQL for example so making this functionality could be possible without building the whole tree.

goretkin:
I have understood that the `ObjectShow` approach is not viable, thank you for explaining it.

In an attempt to separate concerns, let's assume (for the time being) that the contents of a transclusion cannot be edited except from a special view (perhaps the document that contains the "source", or perhaps not even that).

- A document is an object, and displaying the document can depend on other objects. For the purposes of discussion, are the terms `object` and `document` synonyms?
- Is it possible to query the middleware (`anytype-heart`) for one block within a document, including via subscriptions?
- Should it be possible to do the above? Or should the query / subscription granularity remain at the level of objects? If that's the case, then each block itself should be an object. Is that that the ultimate goal?
- Is it clear what the data model for a transclusion block should be, even though the infrastructure may not support it today? Would it be to  introducing a new [`ObjectLayout`](https://github.com/anyproto/anytype-ts/blob/develop/src/ts/interface/object.ts#L1) for transclusion?

ra3orblade:
> I have understood that the `ObjectShow` approach is not viable, thank you for explaining it.
> 
> In an attempt to separate concerns, let's assume (for the time being) that the contents of a transclusion cannot be edited except from a special view (perhaps the document that contains the "source", or perhaps not even that).
> 
> * A document is an object, and displaying the document can depend on other objects. For the purposes of discussion, are the terms `object` and `document` synonyms?
> * Is it possible to query the middleware (`anytype-heart`) for one block within a document, including via subscriptions?
> * Should it be possible to do the above? Or should the query / subscription granularity remain at the level of objects? If that's the case, then each block itself should be an object. Is that that the ultimate goal?
> * Is it clear what the data model for a transclusion block should be, even though the infrastructure may not support it today? Would it be to  introducing a new [`ObjectLayout`](https://github.com/anyproto/anytype-ts/blob/develop/src/ts/interface/object.ts#L1) for transclusion?

1. Well, for the terms of discussion I assume that it's the same. We do not use a term document internally though for me its a document. It's basically a tree of blocks.
2. No, right now it's not possible and there is no subscription for separate blocks. This is planned for implementation in order to make transclusion properly.
3. Yes, read the above
4. Umm, I have no idea yet, we had a discussion on this recently and I do not like the technical proposal yet so we will have more discussions on this.

goretkin:
My 3 above should have been split into two.

To clarify, is the ultimate goal to have each block be its own object?

Is the discussion and technical proposal you mentioned public?

Thanks!

ra3orblade:
No, each block should not become and object. The idea is to have an API to quickly search and subscribe to blocks with SQL-like queries in any-store. No, this discussion was internal with infrastructure lead.

PR #1967, opened 2026-02-10
ci: migrate GitHub workflows to Bun

lubasinkal:
## Summary

This PR updates the GitHub Actions workflows to use Bun instead of npm for faster CI/CD builds.

### Build Workflow (`.github/workflows/build.yml`)
- Replaced `actions/setup-node` with `oven-sh/setup-bun`
- Replaced `npm install --save-dev webpack-cli --legacy-peer-deps` with `bun install`
- Updated all `npm run` commands to `bun run`:
  - `npm run update:locale` → `bun run update:locale`
  - `npm run build:nmh-win` → `bun run build:nmh-win`
  - `npm run build:nmh` → `bun run build:nmh`
  - `npm run release:notes` → `bun run release:notes`

### Nightly Workflow (`.github/workflows/nightly.yml`)
- Same changes as build workflow
- Updated all npm commands to use Bun

## Benefits

- **Faster installs**: Bun is 5-10x faster than npm
- **Simpler setup**: No need for manual webpack-cli installation
- **Consistent**: Aligns with the Bun migration already done in package.json

## Testing

- [ ] Build workflow passes
- [ ] Nightly workflow passes
- [ ] Release artifacts are created correctly

## Related

Part of the Bun migration documented in BUN_MIGRATION.md


lubasinkal:

ra3orblade (changes requested):
## Do NOT merge — same issues as #1968 plus additional CI concerns

This PR contains the exact same `package.json`, `bun.lock`, `bunfig.toml`, and `scripts/` changes as #1968 (shares the same base commit `2146cd42`), with CI workflow changes added on top. All issues from #1968 apply here too:

### 1. Breaks npm users (not backward compatible)
All npm/npx references in `package.json` scripts are replaced with bun/bunx — including `postinstall` (`bunx patch-package`). Anyone without Bun installed cannot even run `npm install`.

### 2. Broken indentation in `package.json`
Multiple lines lost their leading tab:
- `"scripts": {`
- `"doctor:start:dev":`
- `"start:watch":`
- `"dist:mac":`
- `"ext:manifest:firefox":`
- `"build:web":`

### 3. CI workflow changes remove Node.js entirely
- Removes `actions/setup-node` — but Node.js is still needed (Electron, rspack, electron-builder all require Node)
- Uses `bun-version: latest` which is non-deterministic and will break builds when Bun releases breaking changes
- `bun install` does not respect `package-lock.json`, so CI builds become non-reproducible

### 4. Unnecessary scripts with quality issues
- `clean.ts` runs `rm -rf dist/*` — the `dist/` folder contains important non-build files (workers, libs, json) checked into the repo
- `dev-server.ts` duplicates what `npm-run-all --parallel` already does

### 5. No CI validation
Only CLA check and Socket security ran. No build or typecheck was verified.

---

**Recommendation:** Close this PR. If Bun migration is desired, it should:
- Keep npm as the primary/fallback (not replace it)
- Pin Bun version in CI instead of `latest`
- Keep `actions/setup-node` alongside Bun (Node.js is still required)
- Preserve `package-lock.json` for reproducible builds
- Not break `postinstall` for npm users

PR #1998, opened 2026-02-20
perf: replace Array.find/includes with Set/Map for O(1) lookups in loops

Ray0907:
Description

  This PR replaces Array.find() / Array.includes() with Set and Map for O(1) lookups inside
   loops across several hot-path stores and components.

  Key optimizations:
  - Convert arrays to Set before filtering loops (new Set(list.map(...)) + .has() instead
  of Array.includes())
  - Use Map for index-based ordering (new Map(ids.map((id, i) => [id, i]))) instead of
  repeated Array.indexOf()
  - Use Set for deduplication checks in chat message and record group operations

  Files changed:
  - src/ts/store/chat.ts — Set-based dedup for message list operations; Map-based
  reply/state lookups
  - src/ts/store/record.ts — Set/Map for view ordering, group filtering, and relation dedup
  - src/ts/store/common.ts — formatting cleanup
  - src/ts/store/menu.ts — formatting cleanup
  - src/ts/component/block/chat.tsx — Set for scroll tracking; Map for section grouping

PR #1999, opened 2026-02-20
perf: replace Array.find/includes with Set/Map for O(1) lookups in loops

Ray0907:
## Summary

Replace O(n) array lookups (`Array.includes`, `Array.find`, `Array.indexOf`) with O(1) `Set`/`Map` lookups in hot paths where these calls happen inside loops or `.filter()`.

## Changes

**`store/chat.ts`** — `prepend` / `append`: build a `Set` of existing message IDs before filtering duplicates

**`store/record.ts`** — 3 methods:
- `viewsSort`: replace `ids.indexOf()` inside `.sort()` with a pre-built `Map` (O(n^2 log n) → O(n log n))
- `groupsAdd`: replace `list.find()` inside loop with a `Set`
- `groupsRemove`: replace `ids.includes()` inside `.filter()` with a `Set`

**`store/menu.ts`** — `isOpen`: replace `filter.includes()` (called twice per item) with a `Set`

**`component/block/chat.tsx`** — `getSections`: replace `sections.find()` inside `.forEach()` with a `Map` for date-based grouping

## Note

This is a clean re-submission of #1998 with only the semantic changes — no formatting modifications.

### DIFF
--- a/src/ts/component/block/chat.tsx
+++ b/src/ts/component/block/chat.tsx
@@ -329,15 +329,17 @@ const BlockChat = observer(forwardRef<RefProps, I.BlockComponent>((props, ref) =
 	const getSections = () => {
 		const sections = [];
 
+		const sectionMap = new Map();
 		messages.forEach(item => {
 			const key = U.Date.dateWithFormat(I.DateFormat.ShortUS, item.createdAt);
-			const section = sections.find(it => it.key == key);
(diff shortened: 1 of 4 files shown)


ra3orblade (approved):

ra3orblade:
Thanks, now good

PR #2201, opened 2026-05-03
JS-2193: ignore stale ObjectOpen after workspace switch

### LINKED ISSUE #2193: 🐞 Bug Report: Error when switching between workspaces (ObjectOpen: tree does not exist)

### Current Behavior

Describe the bug

When switching between workspaces using keyboard shortcuts (Cmd + 1, Cmd + 2), returning to a previously opened workspace sometimes causes an error:

"There was a problem while opening the object, please send the debug info to the support team."

The object fails to load after switching back.

https://github.com/user-attachments/assets/c512c52e-186e-4565-baf9-298362834e90

**Actual behavior**

Error message appears:
"There was a problem while opening the object..."
Object fails to load

Console shows error:
Error ObjectOpen code: 1 description: build tree: tree does not exist

### Expected Behavior

Objects should open normally when switching between workspaces without errors.

### Steps To Reproduce

Open Anytype
Switch to Workspace 1 (Cmd + 1)
Switch to Workspace 2 (Cmd + 2)
Repeat this in a loop frequently.

### Environment

```markdown
- OS: Mac OS
- Version: 26.4.1 (25E253)
```

### Anything else?

_No response_

Toseef-Ahmad:
## Summary
Fixes [#2193](https://github.com/anyproto/anytype-ts/issues/2193): rapid workspace switching (e.g. Cmd+1 / Cmd+2) could surface **ObjectOpen** errors such as `build tree: tree does not exist` and the generic “problem while opening the object” popup even when the UI had already moved on.

## Root cause
`ObjectClose` from `pageClose` / space switch could finish before an in-flight `ObjectOpen` response returned. The middleware had already torn down the tree, so the late response was treated as a hard failure.

## Approach
- **`C.ObjectOpen`** (`command.ts`): optional `isSuperseded` callback; when true, skip `onObjectView` / last-opened storage so a superseded success cannot apply the wrong tree.
- **Object pages** (editor, set, media, chat, relation, date): per-mount **open sequence** + space snapshot; bump on each `open` / `close`; treat matching responses as stale and skip UI/error handling (loaders still clear).
- **Guard** empty `rootId` where we issue `ObjectOpen`.

## Testing
- [x] Switch spaces quickly with an object open (and while a page is still loading); confirm no spurious error popup.

---
https://github.com/anyproto/.github/blob/main/docs/CLA.md

---

- [ ] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/src/ts/lib/util/router.ts
+++ b/src/ts/lib/util/router.ts
@@ -161,7 +161,12 @@ class UtilRouter {
 			sidebar.rightPanelClose(false, false);
 		};
 
-		if (routeParam.spaceId && (routeParam.spaceId != space) && ![ 'object', 'invite' ].includes(routeParam.action)) {
+		// If route explicitly targets another space, switch before navigating.
+		// NOTE: Universal object routes (`/object?objectId=...&spaceId=...`) are normalized to
+		// `{ page: 'main', action: 'object', id: objectId, spaceId }` in `getParam()`.
+		// They must also trigger a space switch; otherwise we try to open the object in the
+		// previous space (stale `S.Common.space`) when switching spaces quickly.
+		if (routeParam.spaceId && (routeParam.spaceId != space) && ![ 'invite' ].includes(routeParam.action)) {
 			this.switchSpace(routeParam.spaceId, newRoute, false, param, false);
 			return;
 		};


ra3orblade:
I do not think this change should be implemented like this as it might break in some cases and does not look stable. The problem why this popup appears is that we are trying to open object from a different space, so right now it takes space from the store, it might be better to fix it by using space from router in ObjectOpen request.

Toseef-Ahmad:
## Summary
Fixes [#2193](https://github.com/anyproto/anytype-ts/issues/2193): when switching workspaces/spaces quickly, opening an object via a deep link could use a **stale `spaceId`** and attempt to open the object in the previous space, leading to “problem while opening the object” / missing tree errors.

## Root cause
`U.Router.go()` intentionally **skipped** `switchSpace()` for routes with `action: 'object'`. Universal object links like `object?objectId=...&spaceId=...` are normalized to `page: main, action: object` and include a `spaceId`, but the skip meant the app could navigate while `S.Common.space` was still the **previous** space.

## Approach
- **`U.Router.go`** (`src/ts/lib/util/router.ts`): if a route includes `spaceId` and it differs from the current space, **trigger `switchSpace()` for `action: 'object'` as well** (still excluding `invite`).
- This ensures object pages always open after the correct space is active, preventing `C.ObjectOpen(..., S.Common.space)` from using an outdated space.

## Testing
- [x] Switch spaces rapidly (Cmd+1 / Cmd+2) and open objects from deep/universal links (`objectId` + `spaceId`); confirm no stale-space open errors and the correct object loads.

---
`https://github.com/anyproto/.github/blob/main/docs/CLA.md`

Toseef-Ahmad:
@ra3orblade Thanks for taking a look. After your comment I revisited the root cause and simplified the fix down to a single-line change. I cherry-picked the update onto this branch, so you can review the latest commit. Hope this approach makes sense and is worth considering.

ra3orblade (approved):

ra3orblade:
> @ra3orblade Thanks for taking a look. After your comment I revisited the root cause and simplified the fix down to a single-line change. I cherry-picked the update onto this branch, so you can review the latest commit. Hope this approach makes sense and is worth considering.

Thanks! That is much less confusing, thanks for digging into this problem!

PR #2240, opened 2026-06-02
feat(sidebar): add Show as toggle for My Favorites section

yasyuk:
## Summary

- Adds a **"Show as: List / Widgets"** context-menu option to the My Favorites sidebar section
- **List** (default) — existing compact icon + name rows via `WidgetObject`
- **Widgets** — each personal widget rendered as a full `WidgetIndex` card with its own header, collapse, layout settings, and options menu (layout is persisted per-item in the personal widgets block store)
- "Show as" is hidden when the channel's Sidebar View setting is **Links**; in that mode My Favorites always uses the flat list
- Selected view is persisted per-space via `widgetSections` storage

## Bug fixes (post-review)

- **MobX self-assignment** — `updateWidgetSection` was doing `widgetSectionsSet(this.widgetSections)` (same array reference), so MobX detected no change and the sidebar never re-rendered. Fixed with spread: `[ ...this.widgetSections ]`
- **Cold-render section vanish** — `getWidgets()` was filtering personal widget blocks via `S.Detail.get(personalRootId, targetId)` which returns `_empty_` before the async `C.ObjectOpen` resolves, hiding the whole section. Removed the detail-store check; block structure validity is sufficient
- **Stale subscription** — `WidgetObject`'s `useEffect([])` captured target IDs at mount and never re-subscribed when favorites changed. Refactored into `updateData()` exposed via `useImperativeHandle`, which slots into the existing `updateWidgetData` event system that `WidgetIndex` already routes to child components
- **Cross-store DropTarget** — `targetTop`/`targetBot` in `WidgetIndex` were hardcoded to `rootId={S.Block.widgets}`, but personal widget block IDs live in `personalRootId`. Gated behind `!props.rootId` so personal widget blocks get no drop zones
- **Unsafe preview path** — passing `setPreview` to personal widget blocks could trigger a preview lookup in `S.Block.widgets` (wrong store). Now passes `setPreview={undefined}` for `isFavWidgets` blocks

## Test plan

- [ ] Toggle "Show as: Widgets" from My Favorites context menu → section switches to individual widget cards immediately
- [ ] Toggle "Show as: List" → returns to compact flat list
- [ ] Add / remove a favorite while List mode is active → list updates without sidebar reload
- [ ] Switch channel Sidebar View to "Links" → "Show as" absent from context menu; section renders as flat list
- [ ] In Widgets mode, drag a pin-section widget → no drop highlight on personal widget cards
- [ ] Each personal widget card shows its correct layout (Tree, Compact, etc.) and responds to the layout options menu
- [ ] `bun run typecheck && bun run lint` passes clean

### DIFF
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -2408,6 +2408,10 @@
 	"widgetRecentEditMode0": "Anyone",
 	"widgetRecentEditMode1": "Only me",
 
+	"widgetFavoritesViewTitle": "Show as",
+	"widgetFavoritesViewList": "List",
+	"widgetFavoritesViewWidgets": "Widgets",
+
 	"widgetManageSections": "Manage Sections",
 	"widgetHideSection": "Hide section",
 	"widgetShowSection": "Show section",
--- a/src/ts/component/menu/widget.tsx
+++ b/src/ts/component/menu/widget.tsx
@@ -6,8 +6,10 @@ const MenuWidget = forwardRef<I.MenuRef, I.Menu>((props, ref) => {
 
 	const { param, close, setActive, onKeyDown, position, getId, getSize } = props;
 	const { data, className, classNameWrap } = param;
-	const { blockId, isPreview, target } = data;
+	const { blockId, isPreview, target, rootId: blockRootId } = data;
 	const { widgets } = S.Block;
+	const effectiveRootId = blockRootId || widgets;
+	const isPersonalWidget = !!blockRootId;
 	const [ layout, setLayout ] = useState<I.WidgetLayout>(data.layout);
 	const [ limit, setLimit ] = useState(data.limit);
 	const nodeRef = useRef(null);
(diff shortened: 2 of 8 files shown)


yasyuk:
I have read the CLA Document and I hereby sign the CLA.

requilence:
Rebased onto latest `develop` and pushed two follow-up fix commits after a deeper review.

### `a7fc25b6` — harden the Show-as data path
- **Subscription race** — `WidgetObject.updateData()` called `destroyList()` then `subscribe()` synchronously; the async unsubscribe ack could land *after* the new subscribe and clear the just-created subscription, intermittently blanking the favorites list. Nested `subscribe` inside the `destroyList` callback (matches the established `WidgetIndex.getData` pattern).
- **Ghost archived/deleted cards** — in Widgets mode the section filter only checked that a target id existed, so an archived/deleted favorite (whose block still exists) rendered an empty card; List mode already hid these. Added an archived/deleted guard that still tolerates `_empty_` so the section doesn't vanish on cold render.

### `f24ab360` — guard the widget cards
- **Reachable crash** — the "See all" button on a personal-widget card calls `onSetPreview()` → `setPreview(blockId)`, but `setPreview` is intentionally `undefined` for these cards, so clicking it on a favorited Set/page whose item count exceeds the limit threw `setPreview is not a function`. Guarded `onSetPreview`.
- **Dead drag affordance + bogus analytics** — fav cards rendered `draggable` but with no drag handlers and no drop targets (inert), while `onDragEnd` still fired a `ReorderWidget` analytics event on every no-op drag. Gated `draggable` on `!rootId`.
- Minor: undefined guard in `updateWidgetSection`; brace/`==` style nits.

### Verification
Added a Playwright E2E spec in **anytype-desktop-suite** (branch `feat/my-favorites-show-as`) that drives the real app and confirms — on the live page, with no reload — that List↔Widgets toggling and favorite add/remove repaint the section, and that "Show as" is hidden when Sidebar View is Links. 2/2 passing, stable across runs. `typecheck` + `lint` clean.

PR #2249, opened 2026-06-11
Startup performance: chunk splitting, parallel boot, timing marks

requilence:
## What

Startup-performance changes:

**1. Fix eager chunk loading (`manualChunks`)** — the actual win:
- Vite's preload-helper virtual module was colocated into `vendor-mermaid.js`, so `main.js` statically pulled the whole 4.6MB mermaid chunk into every startup. Now routed to a tiny `helpers` chunk.
- The blanket `node_modules → vendor` rule force-merged dynamically-imported packages (katex, @viz-js, react-pdf, pako) into the static `vendor.js`. Now excluded so they keep their own lazy chunks.

**2. Parallelize window creation with middleware startup** — `createWindow()` no longer waits for anytypeHelper's gRPC address; renderer boot overlaps middleware startup. Renderer awaits `Api.getServerAddress` before `dispatcher.init` (fast path kept).

**3. Boot timing marks (`U.Perf`)** — thin User Timing wrapper (`src/ts/lib/util/perf.ts`); `boot:*` milestones across the boot path, visible in the DevTools Performance panel (Timings track).

## Result (production build)

| | Before | After |
|---|---|---|
| Eager startup JS | ~19.5MB | **13.0MB (−33%)** |
| vendor.js | 3.7MB | 1.3MB |
| vendor-mermaid (4.6MB) | eager | lazy |

Plus parallelization saves min(renderer boot, middleware boot) on cold start (~0.5–2s).

## Testing

- `typecheck` + `lint` pass
- Smoke test: embed blocks (mermaid/latex/graphviz), PDF block, graph view (now load vendor code on first use); normal login flow.

### DIFF
--- a/electron/ts/api.ts
+++ b/electron/ts/api.ts
@@ -65,6 +65,15 @@ class Api {
 		};
 	};
 
+	/**
+	 * Resolves with the gRPC web proxy address once the middleware is up.
+	 * Windows are created in parallel with server startup, so the renderer
+	 * awaits this before initializing the dispatcher.
+	 */
+	getServerAddress (win: AppWindow): Promise<string> {
+		return Server.whenReady();
+	};
+
 	logout (win: AppWindow): void {
 		WindowManager.sendToAllTabs('logout');
 	};
--- a/electron/ts/main.ts
+++ b/electron/ts/main.ts
@@ -8,7 +8,7 @@ declare global {
 process.stdout?.on?.('error', () => {});
 process.stderr?.on?.('error', () => {});
 
-import { app, BrowserWindow, session, nativeTheme, ipcMain, powerMonitor, dialog } from 'electron';
+import { app, BrowserWindow, session, nativeTheme, ipcMain, powerMonitor, dialog, contentTracing } from 'electron';
 import { is, fixPathForAsarUnpack } from 'electron-util';
 import path from 'path';
 import storage from 'electron-json-storage';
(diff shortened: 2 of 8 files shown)


requilence:
## Measured A/B — packaged arm64, identical instrumentation

Baseline = `develop` + the same `U.Perf` marks and trace capture, but **without** the chunking fix and parallelization. Both apps built arm64, run back-to-back on the same machine, traced with `ANYTYPE_TRACE_STARTUP=1` (same lean categories).

### Deterministic metrics (independent of run warmth)

| | Before | After |
|---|---|---|
| Eager JS at startup | ~19.5 MB (vendor-mermaid 4.6 MB eager) | **13.0 MB (−33%)** |
| V8 background parse at boot | **626 ms** | **339 / 423 ms** (two runs, −32…46%) |
| vendor-mermaid.js at startup | fetched 193 ms + parsed 179 ms | not loaded — fetched on demand (manually verified working) |
| vendor.js parse | 106 ms | 39 / 59 ms |

### Wall-clock milestones (1 before run, 2 after runs)

| Milestone | Before | After |
|---|---|---|
| Window created | 297 ms | 158 / 234 ms |
| Bundle parsed+evaled (`boot:entry`) | 950 ms | 689 / 904 ms |
| **UI interactive (`boot:ready`)** | **1,127 ms** | **834 / 1,068 ms** |
| Renderer wait for middleware (`boot:init→boot:server`) | n/a — serialized before window creation | 5.5 / 5.7 ms |
| AccountSelect → first command batch (middleware) | 577 ms | 612 / 642 ms — unchanged, as expected |

Run-to-run variance is roughly ±150 ms (disk-cache warmth); the parse-work and bytes numbers are the reliable signal.

### Correction to my earlier estimate
I previously estimated the parallelization saves ~0.6–0.7 s — **that was wrong**. On a warm machine the helper prints its gRPC address in ~100–150 ms (the heavy account/space work happens *after* address print and is unaffected by window timing), so the measured win here is ~100 ms of window-appearance latency. It scales with middleware start time (cold disk, first launch, slower machines benefit more). The **chunking fix is the dominant measured improvement**: −33% eager bytes, −200…290 ms parse work at every startup.

### Remaining startup costs (out of scope here)
- `WorkspaceOpen` ~2.3 s middleware-side (147-space account)
- Main-thread decode stalls on large responses (0.3–0.8 s, dispatcher/mapper)
- main.js parse ~0.2–0.25 s → future route-level code splitting

PR #2259, opened 2026-06-17
fix(comments): split object-page comment menu into "Copy message link" and "Copy link"

requilence:
> **Scope:** this PR changes the **object-page comment system** (`component/comment/` — `post.tsx` / `reply.tsx`), **not** the space chat. The bug reported in [JS-9812](https://linear.app/anytype/issue/JS-9812) is in the **space chat**, fixed by **#2260**. This PR applies the same UX improvement to the comment menu as a bonus.
>
> ⚠️ The branch name `JS-9812-chat-copy-link` is a misnomer — it predates discovering that comments and chat are separate systems.

## Problem
In the object-page comment/reply context menu, a single **Copy link** action always copied the *comment* deeplink (`anytype://…&messageId=…`), regardless of what was right-clicked. Right-clicking a URL in the comment text copied the comment link instead of the URL under the cursor.

## Solution
Split into two distinct actions:
- **Copy message link** — copies the comment deeplink (former behavior, always shown).
- **Copy link** — copies the actual URL under the cursor (`<a href>`); shown in the **Copy text** group only when the menu is opened on a link.

Applied to comments (`post.tsx`) and threaded replies (`reply.tsx`). New i18n key `commentCopyMessageLink`; `commentCopyLink` → "Copy link".

## Tests
E2E coverage (comment menu) in anyproto/anytype-desktop-suite#51.

## Verification
`bun run typecheck` ✓ · `bun run lint` ✓ (0 errors)

### DIFF
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -3009,7 +3009,8 @@
 	"commentDelete": "Delete",
 	"commentSend": "Send",
 	"commentCopyText": "Copy text",
-	"commentCopyLink": "Copy link to comment",
+	"commentCopyLink": "Copy link",
+	"commentCopyMessageLink": "Copy message link",
 	"commentRepliesCount": "%s replies",
 	"commentReplyCount": "%s reply",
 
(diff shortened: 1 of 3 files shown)


requilence:
manually checked

PR #2260, opened 2026-06-17
fix(chat): split space-chat message menu into "Copy message link" and "Copy link" (JS-9812)

requilence:
> ✅ **This is the fix for [JS-9812](https://linear.app/anytype/issue/JS-9812)** — the **space chat** (`component/block/chat/`, menu built in `block/chat.tsx`).
> For the same change applied to the separate **object-page comment** system, see the bonus PR **#2259**.

## Problem
In the **space chat**, the message context menu had a single **Copy Link** action that always copied the *message* deeplink (`anytype://…&messageId=…`), even when right-clicking a URL in the message text — misleading.

## Solution
Split the menu action in `block/chat.tsx`:
- **Copy message link** — copies the message deeplink (former `link` behavior, always shown).
- **Copy link** — copies the actual URL under the cursor; shown in the **Copy text** group only when the menu is opened on a link (`a.markuplink`) in the message text.

URL is detected from the right-click target (`(e.target).closest('a.markuplink')`) in the existing `onContextMenu` handler — no message-component change needed, since chat renders links as `<a href class="markuplink">` and the contextmenu fires on the message element. New i18n keys `blockChatCopyLink` / `blockChatCopyMessageLink`.

## Test plan
- Right-click a plain message → **Copy message link**, no standalone **Copy link**.
- Right-click a URL in a message → **Copy text** + **Copy link** (grouped) and **Copy message link**.
- **Copy link** → URL on clipboard (no `messageId=` deeplink).
- **Copy message link** → `anytype://…messageId=…` deeplink on clipboard.

## Tests
E2E coverage (space-chat menu) in anyproto/anytype-desktop-suite#51.

## Verification
`bun run typecheck` ✓ (no errors in changed files) · `bun run lint` ✓

### DIFF
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -925,6 +925,8 @@
 	"blockChatReplying": "Replying to %s",
 	"blockChatFormReadonly": "Only editors can send messages. Contact the owner to request access.",
 	"blockChatCopyText": "Copy Text",
+	"blockChatCopyLink": "Copy link",
+	"blockChatCopyMessageLink": "Copy message link",
 	"blockChatDownloadFiles": "Download %s files",
 	"blockChatMarkAsUnread": "Mark as unread",
 	"blockChatAddAttachment": "Add attachment",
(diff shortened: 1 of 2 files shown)


requilence:
manually tested

PR #2261, opened 2026-06-17
feat(chat): download single or all files from message context menu (JS-9813)

requilence:
## Problem

In a chat message with **more than one file**, the right-click context menu had **no Download item** at all — it only appeared when a message had exactly one downloadable file. The handler also only ever downloaded the first file (`files[0]`).

Closes JS-9813.

## Solution

The context menu now resolves the **file the user right-clicked on** and shows up to two items:

| Case | Menu |
|---|---|
| 1 file | **Download** |
| Many files, right-clicked a specific file | **Download** (that file) + **Download all** |
| Many files, opened via the "…" more button (no specific file) | **Download all** |

`Action.downloadFiles()` batches a multi-file download through a **single** directory dialog (no N popups).

## Changes

- `src/ts/component/block/chat/attachment/index.tsx` — attachment root carries `data-id` so the clicked file is identifiable.
- `src/ts/component/block/chat.tsx` — `onContextMenu` resolves the clicked file via `closest('.attachment')`; `getMessageMenuOptions` builds the two items; `download` targets the clicked/first file; new `downloadAll` handler.
- `src/ts/lib/action.ts` — new `Action.downloadFiles()` downloads many files through one directory dialog; `downloadFile` delegates to it.
- `src/json/text.json` — added `commonDownloadAll`.

## Testing

- `bun run typecheck` ✅
- `bun run lint` ✅ (biome + eslint clean on changed files)

### DIFF
--- a/src/json/constant.ts
+++ b/src/json/constant.ts
@@ -72,6 +72,7 @@ export default {
 
 		string: {
 			mention:					 300,
+			fileName:					 24,
 		},
 	},
 
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -140,6 +140,7 @@
 	"commonUploadComputer": "Upload from computer",
 	"commonLoad": "Load",
 	"commonDownload": "Download",
+	"commonDownloadFile": "Download %s",
 	"commonDownloadAll": "Download all",
 	"commonDownloading": "Downloading...",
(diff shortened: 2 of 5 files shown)


requilence:
tested manually

PR #2266, opened 2026-06-23
fix(space): stop opening stale objectId from previous space on switch (JS-9815)

requilence:
## Problem (JS-9815)

Intermittently, switching spaces shows the "There was a problem while opening the object" error. The failing `ObjectOpen` carries the **new** spaceId but an **objectId from the previously opened space**, so the middleware returns `build tree: tree does not exist`.

## Root cause

`lastOpenedSimple` is a space-scoped storage key, but both the read and the write resolved the bucket from the globally-mutable `S.Common.space` at call time — and the write happens in the async `ObjectOpen` gRPC callback.

A late `ObjectOpen` response from the previous space, arriving after the user already switched (`S.Common.spaceSet` runs in `onInfo` during the switch), wrote that previous object into the **new** space's bucket. The next switch then read that bucket and tried to open `oldObjectId` in `newSpaceId`. `getLastObject()` made it worse by blindly stamping the current space onto whatever id it read.

## Fix

- **`command.ts`** — key the last-opened write by the object's **actual** space (`object.spaceId`, falling back to the request's `spaceId`, which is captured at call time), not the globally-current space; persist that spaceId in the entry.
- **`storage.ts`** — `getLastOpened`/`setLastOpened` accept an explicit `spaceId` (defaults to current — backward compatible).
- **`space.ts`** — `getLastObject()` validates the entry's space and returns `null` on mismatch instead of blind-stamping.
- **`common.ts`** — `checkErrorOnOpen` self-heals a **stale** last-opened entry (legacy/no-spaceId or foreign-space) when its object fails to open, so buckets polluted before this fix recover. Valid same-space entries are kept (a transient failure must not drop the restore point).

## Review

Reviewed by 3 independent agents (correctness/race, regressions, tests): write-side race confirmed completely closed; storage refactor confirmed behavior-preserving; the self-heal was narrowed per review to avoid over-clearing.

## Tests

New unit tests (vitest), all confirmed fail-before / pass-after where applicable:
- `storage.test.ts` — per-space isolation + the race guard.
- `space.test.ts` — `getLastObject` space validation + legacy passthrough.
- `common.selfHeal.test.ts` — self-heal clears stale, keeps valid, no-op on id mismatch.

typecheck + eslint + biome clean. (The repo's vitest harness has ~101 pre-existing failures from un-injected AutoImport globals — unchanged by this PR.)

### DIFF
--- a/src/ts/lib/api/command.ts
+++ b/src/ts/lib/api/command.ts
@@ -1183,11 +1183,15 @@ export const ObjectOpen = (objectId: string, traceId: string, spaceId: string, c
 		if (!message.error.code) {
 			dispatcher.onObjectView(objectId, traceId, message.objectView, true);
 
-			// Save last opened object
+			// Save last opened object.
+			// Key it by the space the object actually belongs to (not the globally-current
+			// space, which may have already changed if this is a late response after a
+			// space switch) so it never pollutes another space's bucket (JS-9815).
 			const object = S.Detail.get(objectId, objectId, []);
+			const lastSpaceId = object.spaceId || spaceId;
 
 			if (!object._empty_ && ![ I.ObjectLayout.Dashboard ].includes(object.layout) && !keyboard.isPopup()) {
-				Storage.setLastOpened({ id: object.id, layout: object.layout });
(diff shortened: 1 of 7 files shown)


requilence:
@SocketSecurity ignore-all

PR #2305, opened 2026-07-18
fix: restore cursor focus when breaking out of a toggle block (#2287)

### LINKED ISSUE #2287: Cursor loses focus when you double-tap enter to break out of toggle

### Current Behavior

If your cursor is inside a toggle's content, you can hit enter twice to break outside of the toggle. However, this causes the cursor to lose focus and you can't type anything until you use your mouse to click back into it.

I personally find this bug really disruptive as I use this feature all the time for my work.

### Expected Behavior

When you break out of a toggle, the document should not lose focus. This noticeably breaks the flow of the editor.

### Steps To Reproduce

1. Create a toggle
2. Press enter to start typing inside the toggle's body
3. Type some text
4. Press <kbd>enter</kbd>/<kbd>return</kbd> twice to break out of the toggle.

### Environment

```markdown
- OS:Linux
- Version: 0.55.5
```

### Anything else?

It does correctly return the cursor to the same hierarchy as the toggle like it should, but the fact that the editor loses focus means you cannot immediately type anything.

This has been a long-standing issue, not a recent development.

caddydove:
### Issue for this PR

Closes #2287

### Type of change

- [x] Bug fix

### What does this PR do?

Fixes a cursor focus loss when pressing Enter twice to break out of a toggle block.

**Root cause:** When the user presses Enter on an empty child paragraph inside a toggle, `Action.move()` moves the block outside the toggle. The callback (`releaseEnterGuard`) only released the enter processing guard but did **not** re-apply focus on the moved block. Since `Action.move()` triggers a React unmount/remount of the block element, the DOM mutation clears the editor's focus state, and without a `focusSet()` call afterward, the cursor is lost.

**Fix:** The `Action.move()` callback now calls `focusSet(block.id, 0, 0, true)` before releasing the enter guard. This is consistent with how `blockCreate()` and `blockSplit()` handle focus after creating or moving blocks.

### How did you verify your code works?

The fix matches the existing focus restoration pattern used throughout the editor (e.g., `blockCreate()` at line 2680, `blockSplit()` at line 2897). The change is a single callback parameter in one code path.

### Checklist

- [x] I have not included unrelated changes in this PR

### DIFF
--- a/src/ts/component/editor/page.tsx
+++ b/src/ts/component/editor/page.tsx
@@ -1912,7 +1912,10 @@ const EditorPage = forwardRef<I.BlockRef, Props>((props, ref) => {
 			}, releaseEnterGuard);
 		} else
 		if (block.isTextParagraph() && !length && parent && (parent.canToggle() || parent.isTextCallout() || parent.isTextQuote())) {
-			Action.move(rootId, rootId, parent.id, [ block.id ], I.BlockPosition.Bottom, releaseEnterGuard);
+			Action.move(rootId, rootId, parent.id, [ block.id ], I.BlockPosition.Bottom, () => {
+				focusSet(block.id, 0, 0, true);
+				releaseEnterGuard();
+			});
 		} else {
 			blockSplit(block, range, isShift, releaseEnterGuard);
 		};


caddydove:

requilence (approved):

requilence:
Merged, no conflicts. Thanks a lot — clean fix! 🙏

PR #2308, opened 2026-07-21
fix: replace linux-distro with direct /etc/os-release parsing (#2236)

### LINKED ISSUE #2236: App crashes at login with `TypeError: execa is not a function` in `linux-distro`, affects all recent versinos

### Behavior
The app hangs on the login screen. stdout reports:
```
Error occurred in handler for 'Api': TypeError: execa is not a function
    at module.exports (/tmp/.mount_AnytypRJ6NNj/resources/app.asar/node_modules/linux-distro/index.js:11:9)
    at Api.linuxDistro (/tmp/.mount_AnytypRJ6NNj/resources/app.asar/electron.js:3089:18)
    at /tmp/.mount_AnytypRJ6NNj/resources/app.asar/electron.js:3873:31
    at Session.<anonymous> (node:electron/js2c/browser_init:2:116026)
    at Session.emit (node:events:509:28)
{"level":"WARN","ts":"2026-05-29T10:45:09.692+0300","logger":"anytype-grpc","msg":"no servers to broadcast event"}
```

### Cause Analysis
linux-distro uses require('execa') (CommonJS). The version of execa bundled in the app is ESM-only (execa v6+), which does not export a default callable when require()'d — it returns a module namespace object, not a function. This mismatch causes the crash at the call site on line 11 of linux-distro/index.js.
This is a CJS/ESM interop breakage introduced by bundling a modern execa alongside a package that expects the legacy CJS API.

### Expected Behavior

Login completes normally; linuxDistro() resolves without error.

### Steps To Reproduce

Download the latest Anytype AppImage
Make executable and launch
Reach the login screen

### Environment

```markdown
OS: Arch Linux (rolling, up to date)
Anytype version: Reproducible across all recent releases
Install methods: AppImage and Flatpak
Node / Electron: as bundled in AppImage
```

### Anything else?

_No response_

caddydove:
## Description

Fixes #2236 — replaces the `linux-distro` npm dependency with a direct read of `/etc/os-release`, eliminating the fragile `execa` CJS/ESM interop chain that caused `TypeError: execa is not a function` on Linux.

## What changed

**Before:** `require('linux-distro')` which internally called `require('execa')`. Since the bundled `execa` is ESM-only (v7), the CJS `require` resolved a module namespace object instead of a callable function, crashing at the call site on line 11 of `linux-distro/index.js`.

**After:** Read `/etc/os-release` (or `/usr/lib/os-release` / `/etc/lsb-release` as fallbacks) directly with `fs.readFileSync`. Parse the `KEY=value` format and return the `{ os, name, release, code }` shape that `analytics.ts` already expects.

## Why not the previous approach

The earlier commit switched from `require()` to dynamic `import()`. This didn't fix the actual problem — `linux-distro@4` is still a CJS module, and its internal `require('execa')` resolves identically regardless of how we import it from the outside. The `try/catch` only masked the crash while returning a wrong-shaped fallback. This version removes the dependency entirely.

## Verification

- `analytics.ts` reads `data.os` and `data.release` — both are now populated from `/etc/os-release`
- esbuild compilation succeeds
- On non-Linux or missing os-release files, returns `{ os: 'Linux', name: 'Unknown', release: '', code: '' }`

### DIFF
--- a/electron/ts/api.ts
+++ b/electron/ts/api.ts
@@ -669,12 +669,52 @@ class Api {
 		return await checkDiskSpace(app.getPath('userData'));
 	};
 
-	async linuxDistro (win: AppWindow): Promise<{ name: string; version: string }> {
-		const load = require('linux-distro');
-		return await load().catch((err: Error) => {
-			Util.log('error', `[Api].linuxDistro: ${err.message}`);
-			return { name: 'Unknown', version: 'Unknown' };
-		});
+	async linuxDistro (win: AppWindow): Promise<{ os: string; name: string; release: string; code: string }> {
+		try {
+			// Read /etc/os-release (standard on modern Linux) or /etc/lsb-release (older)
+			const paths = [ '/etc/os-release', '/usr/lib/os-release', '/etc/lsb-release' ];
+			let content = '';
+
+			for (const p of paths) {
+				try {
+					content = fs.readFileSync(p, 'utf-8');
+					break;
+				} catch (_) {};
+			};
+
+			if (!content) {
+				return { os: 'Linux', name: 'Unknown', release: '', code: '' };
+			};
+
+			const parse = (text: string): Map<string, string> => {
+				const map = new Map<string, string>();
+				for (const line of text.split('\n')) {
+					const trimmed = line.trim();
+					if (!trimmed || trimmed.startsWith('#')) continue;
+					const eq = trimmed.indexOf('=');
+					if (eq < 0) continue;
+					const key = trimmed.slice(0, eq).trim();
+					let val = trimmed.slice(eq + 1).trim();
+					if ((val.startsWith('"') && val.endsWith('"')) || (val.startsWith("'") && val.endsWith("'"))) {
+						val = val.slice(1, -1);
+					};
+					map.set(key.toLowerCase(), val);
+				};
+				return map;
+			};
+
+			const vars = parse(content);
+
+			return {
+				os: vars.get('id') || vars.get('distrib_id') || 'linux',
+				name: vars.get('pretty_name') || vars.get('name') || vars.get('distrib_description') || '',
+				release: vars.get('version_id') || vars.get('distrib_release') || '',
+				code: vars.get('version_codename') || vars.get('distrib_codename') || '',
+			};
+		} catch (e: any) {
+			Util.log('error', `[Api].linuxDistro: ${e.message}`);
+			return { os: 'Unknown', name: 'Unknown', release: '', code: '' };
+		}
 	};
 
 	menu (win: AppWindow): void {


requilence (changes requested):
Thanks for chasing down the Linux login crash — the crash guard is definitely worth landing. But the diagnosis and the credited mechanism aren't right, and I'd like the framing (and ideally the approach) adjusted before merge.

### What actually fixes the crash
It's the `try/catch`, not the `require → import` switch. `linux-distro@4` is CommonJS and internally does `const execa = require('execa')`, then calls `execa('lsb_release', …)` at line 11 (the crash site). **How we load `linux-distro` from the outside doesn't change its own internal `require('execa')`** — a dynamic `import()` of a CJS module still runs it through the CJS loader, and that internal `require('execa')` resolves identically. So if it was resolving the ESM `execa@7` namespace (non-callable → `execa is not a function`) before, it still is after this change. The only reason the app stops crashing is that the synchronous throw is now caught.

### Why this masks the bug rather than fixing it
With the current patch, `linuxDistro()` silently returns `Unknown/Unknown` on every affected build, and `analytics.ts` (`data.os` / `data.release`) then reports blank distro info. That's better than crashing, but it's graceful degradation, not restored detection.

### Real root cause
An `execa` version mismatch: `linux-distro@4` expects `execa@^0.2.x` (CJS, default-callable) but in the crashing environment resolves the hoisted `execa@7` (ESM, named exports only). Neither `import()` nor the `try/catch` addresses that.

### Suggested approach
- Keep the `try/catch` (good defensive guard) — but rewrite the PR description to credit it, and drop or clearly mark the `require → import` change as incidental.
- Actually fix detection with one of: ensure `linux-distro` resolves a CJS `execa@0.2.x` (nest/pin it), or — simplest and dependency-free — replace `linux-distro` with a direct read of `/etc/os-release` (`ID` / `VERSION_ID`), which removes the fragile `execa` chain entirely.
- Make the fallback shape match what the consumer reads (`{ os, name, release, code }`) so the degraded value is consistent instead of `undefined`.
- Smoke-test on a real `dist:linux` build to confirm esbuild preserves `import()` and the main process executes it.

Happy to help with the `/etc/os-release` version if useful.

requilence:
One more clarification on severity: this isn't actually an app crash — the IPC rejection is already caught in `preload.cjs` (`.catch`) and swallowed in `analytics.ts` (`if (!data) return`), so it's a benign stderr log line on Linux (confirmed by the #2236 reporter who logs in fine with the same error); the only real effect is that distro analytics silently go unrecorded.

caddydove:
You’re right, I conflated preventing the crash with actually restoring distro detection. Dynamic import doesn’t change how `linux-distro` resolves its internal require('execa'), so the current patch only catches the failure and degrades gracefully. Thanks for also catching the return shape mismatch. I’ll rework this to read `/etc/os-release `directly, keep a defensive fallback matching the consumer’s expected shape, update the PR description, and smoke-test the Linux build.

caddydove:
Hi @requilence, I've reworked the PR as you suggested: replaced linux-distro with direct /etc/os-release parsing and fixed the return shape to match { os, name, release, code }. Ready for re-review when you have a moment. Thanks!

PR #2313, opened 2026-07-22
fix: append bookmark URLs to clipboard on copy (#2212)

### LINKED ISSUE #2212: Copying a page with bookmarks is missing the bookmarks

### Current Behavior

When I create a page that contains bookmarks, selecting all text and hitting copy does not add the bookmarks to the clipboard.

Demo:

https://github.com/user-attachments/assets/712b6a2e-aa60-469c-96f9-40f6ad9384dd

This is no macOS Tahoe

### Expected Behavior

I would have expected to get markdown in the pure text representation, or at least have the option to set AnyType up so it exports markdown as the text only export option to the clipboard.

Exporting the page as markdown does seem to work, as in that export contains front-matter and the links as markdown, so the functionality seems to be there, just not connected to the system clipboard.

### Steps To Reproduce

1. Create page
2. add text
3. add bookmark
4. select all
5. copy
6. switch to terminal
7. abpasste

### Environment

```markdown
❯ sw_vers 
ProductName:		macOS
ProductVersion:		26.4.1
BuildVersion:		25E253


anytype 0.54.9
```

### Anything else?

_No response_

caddydove:
## Description

Fixes #2212 — when a user copies blocks that include bookmarks, the bookmark URLs are now included in the clipboard text and HTML content.

## Root cause

The `BlockCopy` gRPC command sends all selected blocks (including bookmarks) to the anytype-heart middleware. The middleware returns three slots:

- `textSlot` — plain text representation
- `htmlSlot` — HTML representation  
- `anySlot` — serialized block objects (for internal Anytype paste)

The middleware does not generate text/HTML content for non-text blocks such as bookmarks. So when pasting into an external app (terminal, text editor), only text block content appears. The `anySlot` already preserves bookmark blocks correctly for internal paste.

## Fix

In `Action.copyBlocks`, after the middleware callback returns, scan the original `blocks` array for `BlockType.Bookmark` blocks that have a URL. Append their URLs to the `textSlot` (newline-separated) and to the `htmlSlot` (as `<a>` links). This ensures bookmark URLs reach the system clipboard in both plain text and HTML formats.

The fix is purely additive — it does not modify or remove any existing clipboard content. If no bookmark blocks are present, the behavior is unchanged.

## Verification

- Unit tests pass (same set of pre-existing failures unrelated to this change)
- When no bookmark blocks are copied, behavior is identical to before
- The `U.String.htmlSpecialChars` helper is used for safe HTML escaping of URLs

### DIFF
--- a/src/ts/lib/action.ts
+++ b/src/ts/lib/action.ts
@@ -679,9 +679,30 @@ class Action {
 				return;
 			};
 
+			let text = String(message.textSlot || '');
+			let html = String(message.htmlSlot || '');
+
+			// The middleware's textSlot/htmlSlot do not include content from
+			// non-text blocks (e.g. bookmarks). Append bookmark URLs so they
+			// appear in the clipboard when pasted into external apps.
+			const bookmarkBlocks = blocks.filter(it =>
+				(it.type == I.BlockType.Bookmark) && it.content?.url
+			);
+			if (bookmarkBlocks.length) {
+				const urls = bookmarkBlocks.map(it => it.content.url);
+				if (text) {
+					text += '\n';
+				};
+				text += urls.join('\n');
+				if (html) {
+					html += '<br/>';
+				};
+				html += urls.map(u => `<a href="${U.String.htmlSpecialChars(u)}">${U.String.htmlSpecialChars(u)}</a>`).join('<br/>');
+			};
+
 			U.Common.clipboardCopy({
-				text: message.textSlot,
-				html: message.htmlSlot,
+				text,
+				html,
 				anytype: {
 					range,
 					blocks: (message.anySlot || []).map(Mapper.From.Block),


caddydove:
**Note:** This fix has a known limitation I want to highlight.

The bookmark URLs are appended to the end of the text/html content. This does not preserve the original block order. For example, if the page layout is the text "paragraph A" followed by a bookmark to "https://example.com" followed by "paragraph B", the clipboard output will be:

```
paragraph Aparagraph Bhttps://example.com
```

This is because the middleware's textSlot/htmlSlot response is an opaque flat string. It does not include position markers for individual blocks. We cannot know where each bookmark was positioned relative to the text blocks.

A proper fix would require changes in the anytype-heart middleware to include bookmark content in textSlot at the correct position. This frontend fix is a pragmatic improvement: it is better to have the URL at the wrong position than to not have it at all.

caddydove:
Hi @requilence, this PR fixes #2212 — bookmark URLs are now included when copying blocks to clipboard. There's a known limitation noted in the PR about URL position ordering. Would appreciate a review when you get a chance. Thanks!

PR #2336, opened 2026-08-12
feat(dataview): add Split master-detail layout for sets and collections

mattaylor:
## What

Adds **Split**, a seventh dataview layout for sets and collections: the record list stays
in a resizable left pane while the selected record's full page renders on the right.
Selecting another record swaps the right pane without navigating, so reviewing many
objects no longer means a repeated open → read → back loop that loses list position.

An expand control in the detail pane opens the current record as a normal full page, so
the split is an entry point to the standard editor rather than a dead end.

Gated behind `config.experimental`, matching the Timeline precedent (`4dca3ca073`).

## Why

Opening a record from a collection always navigates away — via route, popup, or new tab.
For triage-style work across a collection, that context loss is the dominant cost.

## How it works

```
.viewContent.viewSplit          // sized to viewport, own scroll per pane
  .splitMaster                  // renders <ViewList isInline>
  .resize-h > .resize-handle    // drag to resize, width persisted locally
  .splitDetail
    .icon.splitDetailExpand     // floats top-right, over the editor's own title
    .splitDetailBody            // <EditorPage> for the selected object
```

Three constraints shaped the implementation:

**`U.Dom.getScrollContainer` is global.** It resolves to a single element per namespace
(`#page.isFull` / `#page.isPopup`, `lib/util/dom.ts:167`). A detail pane that scrolled the
document would fight the host collection for the same container. So the split container
measures the page container and sizes itself to the viewport, giving each pane its own
internal scroll — the technique `ViewGraph` already uses.

**The master pane reuses `ViewList` unchanged.** It is passed `isInline: true` so it takes
the non-`WindowScroller` branch (`view/list.tsx:59`) and scrolls inside the pane. `isInline`
is threaded only into the master child, never the dataview root, so none of the root's
~20 `isInline` branches change behaviour.

**`EditorPage` had never been embedded** outside `page/main/edit.tsx`. Only the
`I.PageComponent` contract is passed; `S.Common.refSet('editor' + ns)` is deliberately
skipped, since that ref is keyed only by `isPopup` and belongs to the host page's editor.

Split-specific UI state (pane width, last selected record) persists **locally** via
`Lib/storage`, keyed by `rootId` + `blockId` + `viewId`. Nothing new is sent to
middleware beyond the view `type`.

### Middleware compatibility

`Mapper.To.View` sends `type` as a raw number. The checked-in schema at
`dist/lib/protos/models.proto` defines `View.Type` as `Table=0 … Graph=5` — it stops at 5,
so shipped `Timeline=6` *already* exceeds it and round-trips fine, because proto3 open
enums carry unknown numeric values through unchanged. `Split=7` sits in exactly that
position and needs no fallback. Confirmed at runtime: reopening a collection preserves the
Split view.

## Behaviour notes

- **Row click** selects into the detail pane, including the empty space right of the title.
  Cmd-, Shift-, Alt- and middle-click keep their existing open-in-tab/window behaviour;
  right-click passes through to the existing context menu untouched.
- **Keyboard**: ArrowUp/ArrowDown move selection, Enter expands to a full page. Guarded by
  `keyboard.isFocused` so keys are never stolen from the editor in the right pane.
- **Selection is reconciled** against `getRecords()` each render, so a filtered-out or
  deleted record falls back to the first record rather than dangling.
- **Object opening is debounced** (100ms) so fast arrow-key navigation doesn't open and
  close a dozen subscriptions; the pending open is cancelled on unmount. Closing is guarded
  so the detail pane can never close the host collection itself.
- **The add-cover / show-description toolbar is hidden** in the pane (the object is reachable
  as a full page via expand), following the `editorWrapper.isBookmark` precedent.
- **The embedded editor fills the pane.** `EditorPage.getWidth` caps its inline width at
  `J.Size.editor` (704px), which left a dead column beside the text in a wider pane. One
  consequence: the object's own layout-width slider is inert inside the pane — it is
  unreachable there anyway (`#editorSize` is hidden until the block-layout menu opens), and
  pane width is governed by the split resizer instead.

## Scope note for reviewers

**This touches one existing shared component.** `list/row.tsx` hardcoded
`U.Object.openConfig` on row click with no seam to override it, so clicking a row in the
split view navigated away instead of selecting. Fixed by adding an optional

```ts
onRecordClick?(e: any, id: string): boolean;
```

to `I.ViewComponent`, which `ListRow` honours for unmodified left clicks. It is additive
and opt-in — every existing view passes nothing and behaves exactly as before. It has to be
checked in *both* the row handler and the per-cell handler, because the latter calls
`stopPropagation()` and would otherwise bypass the row.

The rest of the change is new files plus additive registration (enum value, dispatch case,
icon map, experimental gate, storage helpers, one translation key).

## Test plan

Automated:
- `bun run typecheck` — clean, both tsconfigs
- `bun run lint` — 0 errors (pre-existing warnings only, in untouched files)
- `bunx vitest run src/ts/model/view.test.ts` — 15/15, extended for `isSplit()`
- Full `bunx vitest run` matches the pre-change baseline exactly (601 passed; the 90
  failures are a pre-existing `ReferenceError: U is not defined` in `lib/util/router.ts:94`,
  unrelated to this branch)
- Dark mode audited: no hardcoded colours, only `--color-shape-*` / `--color-text-*`
  variables that the dark theme already redefines, so no new overrides are needed

Manual, with the experimental flag on:
- Switch a collection to Split; master + detail render
- Navigate away and reopen — view is still Split (middleware round-trip)
- Rapid selection changes; detail pane tracks, editing in the right pane saves, no
  subscription leak or console errors
- Expand → lands on the normal full page for that object
- Drag the resizer, reload, width persisted
- Empty collection, deleted selected record, narrow window, dark mode

**Not covered: E2E.** The Playwright suite lives in a separate repo that was not available
in this working copy, so no automated E2E was added, and the checks that run on PRs here are
CLA and Socket Security rather than a test suite. Flagging rather than omitting silently —
happy to add coverage if you point me at the expected specs and how they are run.

## Follow-up (not in this PR)

Grid as the master pane. `ViewGrid.resize()` measures the page container and applies
negative breakout margins in both branches, and its inline branch looks for `#editorWrapper`,
which doesn't exist inside a pane. Doing it properly needs a `getContainerNode` seam on
`I.ViewComponent`, a `splitMasterType` field with explicit clamping (`Grid === 0` is
swallowed by the `Number(x) || default` idiom), a List/Grid sub-setting in the layout menu,
and local persistence for it. Kept separate to keep this PR reviewable; List is the safer
default regardless.

### DIFF
--- a/docs/src/ts/component/block/README.md
+++ b/docs/src/ts/component/block/README.md
@@ -36,6 +36,7 @@ Database-like component with multiple views. Supports filtering, sorting, groupi
 - `list.tsx` — List with `list/row.tsx`
 - `graph.tsx` — Graph view (uses GraphProvider)
 - `timeline.tsx` — Timeline view
+- `split.tsx` — Master-detail split: `list.tsx` in a resizable left pane, the selected record's `EditorPage` on the right. Sizes itself to the viewport so each pane scrolls internally; pane width and selection persist locally via `Lib/storage` (experimental)
 
 ### Table (`table/` — main file ~1,990 lines)
 Spreadsheet-style tables with `cell.tsx` and `row.tsx`.
--- a/src/json/size.ts
+++ b/src/json/size.ts
@@ -57,6 +57,10 @@ export default {
 			small: 70,
 			medium: 120,
 		},
+
+		split: {
+			master: { min: 240, max: 640, default: 320 },
+		},
 	},
 
 	store: {
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -470,6 +470,7 @@
 	"viewName4": "Calendar",
 	"viewName5": "Graph",
 	"viewName6": "Timeline",
+	"viewName7": "Split",
 
 	"progressDropFiles": "Copying files...",
 	"progressImport": "Import in progress...",
@@ -843,6 +844,8 @@
 	"blockDataviewEmptyTargetButton": "Select Source",
 
 	"blockDataviewEmptyViewDescription": "Create your first one to begin",
+
+	"blockDataviewSplitEmptyDetail": "Select an object to open it here",
 
 	"blockDataviewControlsFilters": "Filters",
 	"blockDataviewControlsSorts": "Sorts",
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -846,6 +846,7 @@
 	"blockDataviewEmptyViewDescription": "Create your first one to begin",
 
 	"blockDataviewSplitEmptyDetail": "Select an object to open it here",
+	"blockDataviewNestedInSplit": "This collection can only be viewed in full screen mode",
 
 	"blockDataviewControlsFilters": "Filters",
 	"blockDataviewControlsSorts": "Sorts",
--- a/src/json/text.json
+++ b/src/json/text.json
@@ -846,6 +846,7 @@
 	"blockDataviewEmptyViewDescription": "Create your first one to begin",
 
 	"blockDataviewSplitEmptyDetail": "Select an object to open it here",
+	"blockDataviewSplitChatDetail": "Chats can only be opened as a full page",
 	"blockDataviewNestedInSplit": "This collection can only be viewed in full screen mode",
 
 	"blockDataviewControlsFilters": "Filters",
--- a/src/scss/block/common.scss
+++ b/src/scss/block/common.scss
@@ -136,6 +136,7 @@
 @import "./dataview/view/calendar";
 @import "./dataview/view/graph";
 @import "./dataview/view/timeline";
+@import "./dataview/view/split";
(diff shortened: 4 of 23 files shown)


mattaylor:

mattaylor:
recheck

mattaylor:
Note - the detail pane is not editable currently when rendered within an inline collection. There is no simple fix for this as `focus` and the selection provider are process-wide singletons that assume one editor per route. Making them editor-aware means threading a rootId through both — core shared code touched by every block, well beyond this PR's footprint. Detail panel editing from a regular collection does work.

mattaylor:
@requilence

PR #2337, opened 2026-08-12
fix(auth): restore session token after renderer reload

cjthevillain:
## Summary

- keep each tab's live session token in runtime tab state
- return that token from `getTab` so a renderer reload can resume the existing session
- keep authentication tokens out of serializable tab metadata
- align the web Electron mock and add regression coverage for the tab contract

## Root cause

`AuthStore.tokenSet()` sends the session token through `updateTab`. The main process merged that update into `view.data`, while `getTab` returned only `{ id, data }`. The renderer expects `tab.token`, so after `Ctrl+R` it could not find the live token and unnecessarily fell back to the OS keychain/recovery flow.

This was reproduced on CachyOS with Anytype 0.56.1. Windows was not tested.

## Validation

- `bun run test -- src/ts/lib/web/electronMock.test.ts`
- `bun run typecheck`
- `bun run lint`
- `bun run build:electron`

The full local test command still reports unrelated existing test-harness failures where auto-imported globals such as `U` and `Relation` are unavailable; the new regression test passes independently.

### DIFF
--- a/electron/ts/api.ts
+++ b/electron/ts/api.ts
@@ -14,7 +14,7 @@ import UpdateManager from './update';
 import Server from './server';
 import Util from './util';
 import { getSafeStorage } from './safeStorage';
-import { AppWindow, TabView, TabData, CreateTabOptions, AppConfig, Bounds } from './types';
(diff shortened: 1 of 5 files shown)


cjthevillain:

PR #2362, opened 2026-09-01
GO-7387: land new objects in their sorted position on commit

brennacodes:

### Description

This PR fixes new objects not landing in their sorted position in name-sorted
Sets and Type views (GO-7387), and renders an empty object name in the
placeholder color so it reads as empty rather than as a real value.

- In a name-sorted dataview, a newly created object was moved to the creation
  spot while its name was typed, then either stayed at the bottom or briefly
  jumped to the top before settling. The position lock is now held through
  commit so the middleware reposition is applied once, moving the row to its
  sorted slot in a single step. Unnamed or unedited records restore their
  stashed position when editing ends.
- A name cell with no name showed the "Untitled" fallback in the normal text
  color, so an empty value looked like a real one. It now renders in the
  placeholder color (`--color-text-tertiary`). A name literally typed as
  "Untitled" keeps the normal color.

- [x] 🐛 Bug Fix
- [x] 🎨 Style

GO-7387 (builds on the earlier collection-sorting fix, PR #2310)

_To add: before/after of the empty-name cell color, and a clip of create-and-sort with no flash._

### Added tests?

- [x] 👍 yes

### Added to documentation?

None.

### DIFF
--- a/src/scss/component/cell.scss
+++ b/src/scss/component/cell.scss
@@ -4,6 +4,10 @@
 	.name { @include text-overflow-nw; display: inline-block; vertical-align: top; width: 100%; }
 }
 
+.cell {
+	.name.isPlaceholder { color: var(--color-text-tertiary); }
+}
+
 .cell.isEditing { border-radius: 2px; }
 .cell.isEditing, .cellContent.isEditing {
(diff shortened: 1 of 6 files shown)


brennacodes:

